PR #20, opened 2017-03-10
Using the correct syntax

### LINKED ISSUE #19: functions-samples uses the $postid instead of {postid} notation for reference string

idanb11:
closes #19


nicolasgarnier:
Thanks @idanb11

PR #25, opened 2017-03-12
Fix/convert use spawn

ooooooo-q:
```
exec(`convert ${fileName}`)
```

exec is execute with shell. And The file name can contain $ {} ".
ex `test$(echo 2).gif`

use spawn or execFile.


ooooooo-q:
I signed it!

nicolasgarnier:
Thanks @ooooooo-q looks good :)

PR #31, opened 2017-03-12
Use transaction() instead of set() in the child-count sample

jwngr:
Using `set()` can lead to race conditions where two children are added simultaneously and each set the value to N+1, instead of the first setting it to N+1 and the second setting it to N+2. Using `transaction()` properly gets rid of that race condition.

See [this Twitter thread](https://twitter.com/_jwngr/status/840063426151960577) for details.

cc/ @puf @aputinski - FYI

### DIFF
--- a/child-count/functions/index.js
+++ b/child-count/functions/index.js
@@ -21,5 +21,8 @@ admin.initializeApp(functions.config().firebase);
 
 // Keeps track of the length of the 'likes' child list in a separate attribute.
 exports.countlikes = functions.database.ref('/posts/{postid}/likes').onWrite(event => {
-  return event.data.ref.parent.child('likes_count').set(event.data.numChildren());
+  var likesCountRef = event.data.ref.parent.child('likes_count');
+  return likesCountRef.transaction(function(currentCount) {
+    return (currentCount || 0) + 1;
+  });
 });


child-count/functions/index.js
```diff
@@ -21,5 +21,8 @@ admin.initializeApp(functions.config().firebase);
 
 // Keeps track of the length of the 'likes' child list in a separate attribute.
 exports.countlikes = functions.database.ref('/posts/{postid}/likes').onWrite(event => {
-  return event.data.ref.parent.child('likes_count').set(event.data.numChildren());
+  var likesCountRef = event.data.ref.parent.child('likes_count');
+  return likesCountRef.transaction(function(currentCount) {
+    return (currentCount || 0) + 1;
```
jamesdaniels: Won't this increment on deletion and update as well?

nicolasgarnier:
Thanks for this @jwngr indeed transactions are way better in this case. Although closing in favor of #32 :) (because it handles decrements)
Too many Firebaser working on weekends :D

child-count/functions/index.js
jwngr: Yup that was an oversight on my part. Thankfully @puf had my back and superseded this PR with #32.

Issue #43, opened 2017-03-14
Uppercase function test undefined input


Cshion:
Hi guys,  thanks for the examples , i am trying to get started with cloud functions but when i want to test uppercase function example there is a weird behavior , {pushId} param from test is not being correctly passing to the function , here is the logs :

the database path from test is : messages/1111/original

```
  Cloud Functions
    makeUpperCase
Uppercasing undefined input
      ✓ should upper case input and write it to /uppercase
    addMessage
      ✓ should return a 303 redirect

```
As you can see, there is a "undefined" in uppercase function , i checked event.params then the result was a empty object {}.Im just curious about this behavior I will try to debug.

katowulf:
HI Aaron, so clearly following the samples to a T works for us. Could you include a minimal repro and some sample data that reproduces the error? Be sure to include version info.

Cshion:
Hi Kato , I tested in node version 4.7 and 6.9 , the log is the same as I showed above , the code was the [uppercase example](https://github.com/firebase/functions-samples/tree/master/quickstarts/uppercase)

katowulf:
Does this happen on every execution or only following a deploy?

Cshion:
it happens when I tested the function ( before to deploy ) , running npm test.

laurenzlong:
Hey @Cshion that's actually expected behavior since we did not mock `params` when constructing the fakeEvent (since our test does not use it). Setting path in the DeltaSnapshot constructor does not actually add a params field.

If you would like to mock it, you can modify the fakeEvent to:

```
js
      const fakeEvent = {
        data: new functions.database.DeltaSnapshot(null, null, null, 'input', 'messages/1111/original'),
        params: {
          pushId: '1111'
        }
      };
```

This will give you the expected console log of : 
Uppercasing 1111 input

Cshion:
it works , thanks!.

laurenzlong:
Not a problem!

PR #52, opened 2017-03-15
Omit path in DeltaSnapshot constructor in unit test to reduce confusion

laurenzlong:
https://github.com/firebase/functions-samples/issues/43

Having a path in the DeltaSnapshot constructor was confusing since it seemed like it would populate the params field of Event. But it doesn't actually do anything, so it's better to not have it.


nicolasgarnier:
We'll need to ping @egilmorez so that he re-publishes the pages that include the code (there is an update planned soon so this time we should be good :))

PR #59, opened 2017-03-17
add payment source

rldaulton:
In the current Firebase Cloud Function example, users lack the ability to add a payment source that they can later initiate a charge on

### DIFF
--- a/stripe/functions/index.js
+++ b/stripe/functions/index.js
@@ -64,14 +64,14 @@ exports.createStripeCustomer = functions.auth.user().onCreate(event => {
 });
 
 // Add a payment source (card) for a user by writing a stripe payment source token to Realtime database
-exports.addPaymentSource = functions.database.ref('/stripe_customers/{userId}/cards/{token}').onWrite(event => {
-  const token = event.data.val();
+exports.addPaymentSource = functions.database.ref('/stripe_customers/{userId}/cards/{pushId}/token').onWrite(event => {
+  const source = event.data.val();
   const user = event.params.userId;
   return stripe.customers.createSource(
     user,
-    {source: token}
+    {source}
   ).then(response => {
-        return console.log('payment source added successfully for user', user);
+        return event.data.ref.parent.set(response);
   }, error => {
       return event.data.ref.child('error').set(userFacingMessage(error)).then(() => {
         return reportError(error, {user: user});
--- a/stripe/functions/index.js
+++ b/stripe/functions/index.js
@@ -63,6 +63,22 @@ exports.createStripeCustomer = functions.auth.user().onCreate(event => {
   });
 });
 
+// Add a payment source (card) for a user by writing a stripe payment source token to Realtime database
+exports.addPaymentSource = functions.database.ref('/stripe_customers/{userId}/cards/{token}').onWrite(event => {
+  const token = event.data.val();
+  const user = event.params.userId;
+  return stripe.customers.createSource(
+    user,
+    {source: token}
+  ).then(response => {
+        return console.log('payment source added successfully for user', user);
+  }, error => {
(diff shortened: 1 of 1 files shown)


rldaulton:
Signed!

stripe/functions/index.js
```diff
@@ -63,6 +63,22 @@ exports.createStripeCustomer = functions.auth.user().onCreate(event => {
   });
 });
 
+// Add a payment source (card) for a user by writing a stripe payment source token to Realtime database
+exports.addPaymentSource = functions.database.ref('/stripe_customers/{userId}/cards/{token}').onWrite(event => {
```
jamesdaniels: what if we listened instead to `/stripe_customers/{userId}/cards/{pushid}/token`? or are you writing token:token to the RTDB. Guess I'm a bit confused on the schema.

stripe/functions/index.js
```diff
@@ -63,6 +63,22 @@ exports.createStripeCustomer = functions.auth.user().onCreate(event => {
...
+  const token = event.data.val();
+  const user = event.params.userId;
+  return stripe.customers.createSource(
+    user,
+    {source: token}
+  ).then(response => {
+        return console.log('payment source added successfully for user', user);
```
jamesdaniels: what if instead we wrote back the response `return event.data.ref.parent.set(response)`, that's what I am doing in my scratch project then you have VISA ***1234, exp date, etc. to allow selection

stripe/functions/index.js
jamesdaniels: we'd also need to change how the stripe customer id was stored

stripe/functions/index.js
```diff
@@ -63,6 +63,22 @@ exports.createStripeCustomer = functions.auth.user().onCreate(event => {
   });
 });
 
+// Add a payment source (card) for a user by writing a stripe payment source token to Realtime database
+exports.addPaymentSource = functions.database.ref('/stripe_customers/{userId}/cards/{token}').onWrite(event => {
+  const token = event.data.val();
+  const user = event.params.userId;
```
jamesdaniels: that isn't going to be the stripe customer id but instead the user id in firebase, but as I mentioned above this schema change conflicts with how the customer_id is currently stored

stripe/functions/index.js
```diff
@@ -63,6 +63,22 @@ exports.createStripeCustomer = functions.auth.user().onCreate(event => {
...
+  return stripe.customers.createSource(
+    user,
+    {source: token}
```
jamesdaniels: nit, if we called token source instead we could use shorthand `{source}`

stripe/functions/index.js
rldaulton: @jamesdaniels yea you're right, I missed one there. Should be `/stripe_customers/{userId}/cards/token`. I'll make this change.

I was going off of what was returned in [customer creation](https://github.com/firebase/functions-samples/blob/e9490d7e8df49a53eb9da44e5865044dc7f90aba/stripe/functions/index.js#L62). Seems to me you are setting the Stripe customer id at this location?

stripe/functions/index.js
rldaulton: [Isn't this the location](https://github.com/firebase/functions-samples/blob/e9490d7e8df49a53eb9da44e5865044dc7f90aba/stripe/functions/index.js#L68) in your schema where you are currently storing a Stripe customer id?

stripe/functions/index.js
rldaulton: Your original suggestion of listening at `/stripe_customers/{userId}/cards/{pushid}/token` is better. Then writing back makes more sense.

rldaulton:
@jamesdaniels [made some changes](https://github.com/firebase/functions-samples/pull/59/commits/258bb4811f7eddd91d6008fe915a9dae6722a862)

PR #71, opened 2017-03-20
Fix mistaken references to "instagram" in linkedin-auth

rjhuijsman:
It looks like a previous revision of the linkedin-auth Cloud Functions sample used Instagram instead of LinkedIn; there were a few references to "instagram" still hanging around the code.


nicolasgarnier:
Thanks @rjhuijsman 
The reason was because we also have an [Instagram sample](https://github.com/firebase/functions-samples/tree/master/instagram-auth) and this one was built on a copy-paste of the other one :D

PR #74, opened 2017-03-21
Switch $ to the correct format - {}

idanb11:


nicolasgarnier:
Thanks @idanb11 looks good!

PR #76, opened 2017-03-21
Username password auth

beaspider:
I have a use case where I need to authenticate with a username/password from a 3rd party system.
I have added an example username-password-auth


beaspider:
Have signed the CLA :)

username-password-auth/main.css
```diff
@@ -0,0 +1,62 @@
+/**
+ * Copyright 2015 Google Inc. All Rights Reserved.
```
nicolasgarnier: I don't think this file is used. remove it? The one that is being served is the main.css in the `public` directory.

username-password-auth/public/main.js
```diff
@@ -0,0 +1,112 @@
...
+  } else {
+    this.signedOutCard.style.display = 'block';
+    this.signedInCard.style.display = 'none';
+  }
+};
+
+// Initiates the sign-in flow.
+Demo.prototype.signIn = function() {
+    var err = this.signInError;
+    err.innerText = "";
```
nicolasgarnier: Can you use single quotes for Strings in JavaScrip in the whole file?

beaspider:
I have completed all the requested changes + added some extra error handling :)

nicolasgarnier:
Looks good, thanks so much @beaspider !

PR #81, opened 2017-03-23
Add Payment Source

rldaulton:
Update from PR #59

- `ref` changed to `/stripe_customers/{userId}/cards/{pushId}/token`. Users can now write back to RTDB and store redacted card information. See line 74: `return event.data.ref.parent.set(response);`
- shorthand on `{source}`

### DIFF
--- a/stripe/functions/index.js
+++ b/stripe/functions/index.js
@@ -32,7 +32,7 @@ exports.createStripeCharge = functions.database.ref('/users/{userId}/charges/{id
   // noop if the charge was deleted, errored out, or the Stripe API returned a result (id exists) 
   if (val === null || val.id || val.error) return null;
   // Look up the Stripe customer id written in createStripeCustomer
-  return admin.database().ref(`stripe_customers/${event.params.userId}`).once('value').then(snapshot => {
+  return admin.database().ref(`stripe_customers/${event.params.userId}/customer_id`).once('value').then(snapshot => {
     return snapshot.val();
   }).then(customer => {
     // Create a charge using the pushId as the idempotency key, protecting against double charges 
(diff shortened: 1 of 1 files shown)


stripe/functions/index.js
```diff
@@ -63,6 +63,22 @@ exports.createStripeCustomer = functions.auth.user().onCreate(event => {
   });
 });
 
+// Add a payment source (card) for a user by writing a stripe payment source token to Realtime database
+exports.addPaymentSource = functions.database.ref('/stripe_customers/{userId}/cards/{pushId}/token').onWrite(event => {
```
jamesdaniels: now that you're storing `cards/**/*` under `stripe_customers/*` you'll need to change where the stripe id for the customer is stored, as currently that's key/value

stripe/functions/index.js
```diff
@@ -63,6 +63,22 @@ exports.createStripeCustomer = functions.auth.user().onCreate(event => {
...
+  const source = event.data.val();
+  const user = event.params.userId;
```
jamesdaniels: You'll then need to lookup that stripe customer id

stripe/functions/index.js
rldaulton: At [this line](https://github.com/rldaulton/functions-samples/blob/916071aa75bf1a9d75c0c6c2e1a0e7956e9b881f/stripe/functions/index.js#L35), correct?

jamesdaniels: Yeah.

    ref(`/stripe_customers/${data.uid}/customer_id`)

Maybe?

rldaulton:
@jamesdaniels [take a look](https://github.com/firebase/functions-samples/pull/81/commits/41b837a604895410cf3b4a324251708ea0e869ad)

PR #84, opened 2017-03-24
createCustomToken returns a promise

mrp14:


mrp14:
I signed it !

instagram-auth/functions/index.js
```diff
@@ -147,8 +147,9 @@ function createFirebaseAccount(instagramID, displayName, photoURL, accessToken)
   // Wait for all async task to complete then generate and return a custom auth token.
   return Promise.all([userCreationTask, databaseTask]).then(() => {
     // Create a Firebase custom auth token.
-    const token = admin.auth().createCustomToken(uid);
-    console.log('Created Custom token for UID "', uid, '" Token:', token);
-    return token;
+    return admin.auth().createCustomToken(uid).then(function (token) {
```
prescottprue: I think fat arrows are preferred: `.then((token) => {`

mrp14: Yes you are right, thanks !

nicolasgarnier:
Good catch thanks! (PS I think the current implementation works but we're moving to Promises as token creation may become async in some cases).

PR #95, opened 2017-03-29
Fix promise chain in count-sample

### LINKED ISSUE #94: Race condition in child-count sample
### How to reproduce these conditions

**Sample name or URL where you found the bug**
https://github.com/firebase/functions-samples/blob/master/child-count/functions/index.js#L42

**Failing Function code used (including require/import commands at the top)**
Using the sample code as is.

**Steps to set up and reproduce**
Using the sample code as is.

**Sample data pasted or attached as JSON (not an image)**
not needed

**Security rules used**
not needed

### Debug output
n/a

**Screenshots**
not needed

### Expected behavior
n/a

### Actual behavior
See my answer here: http://stackoverflow.com/a/43102002/394010

You are returning the once() function, not a promise chain.

Essentially, we've written this:

```
//returns the once promise
return ref.once('value', function(snap) { 
  /* some stuff that doesn't affect the promise chain happens here */
});
```

When what we wanted is this:

```
// returns whatever happens in the .then() callback
return ref.once('value').then(snap => {
   /* now whatever you return here affects the promise chain */
});
```

katowulf:
Fixes #94

### DIFF
--- a/child-count/functions/index.js
+++ b/child-count/functions/index.js
@@ -21,8 +21,8 @@ admin.initializeApp(functions.config().firebase);
 
 // Keeps track of the length of the 'likes' child list in a separate property.
 exports.countlikechange = functions.database.ref('/posts/{postid}/likes/{likeid}').onWrite(event => {
-  var collectionRef = event.data.ref.parent;
-  var countRef = collectionRef.parent.child('likes_count');
+  const collectionRef = event.data.ref.parent;
+  const countRef = collectionRef.parent.child('likes_count');
 
   // return the promise from .transaction() so our function 
   // waits for this async event to complete before it exits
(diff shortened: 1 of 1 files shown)


child-count/functions/index.js
```diff
@@ -39,7 +39,7 @@ exports.recountlikes = functions.database.ref("/posts/{postid}/likes_count").onW
   if (!event.data.exists()) {
     var counterRef = event.data.ref;
     var collectionRef = counterRef.parent.child('likes');
-    return collectionRef.once('value', function(messagesData) {
+    return collectionRef.once('value').then(function(messagesData) {
```
CodingDoug: More idiomatic for ES6:
.then(messagesData => {

child-count/functions/index.js
```diff
@@ -35,11 +35,11 @@ exports.countlikechange = functions.database.ref("/posts/{postid}/likes/{likeid}
 });
 
 // If the number of likes gets deleted, recount the number of likes
-exports.recountlikes = functions.database.ref("/posts/{postid}/likes_count").onWrite((event) => {
+exports.recountlikes = functions.database.ref('/posts/{postid}/likes_count').onWrite((event) => {
```
nicolasgarnier: Remove parenthesis arround `event`
(Same above)

puf:
Ouch. Good catch!

PR #98, opened 2017-03-30
createCustomToken returns a Promise

mrp14:
Same as https://github.com/firebase/functions-samples/pull/84 but for LinkedIn Auth.


nicolasgarnier:
:+1:

PR #102, opened 2017-04-02
Prevent timing attacks on keys / signatures

### LINKED ISSUE #100: Timing Attacks on cron.key
https://github.com/firebase/functions-samples/blob/master/delete-unused-accounts-cron/functions/index.js#L36 compares two keys with !==

I think this might be vulnerable to timing attacks. What about using https://www.npmjs.com/package/secure-compare instead?

Same thing here: https://github.com/firebase/functions-samples/blob/master/github-to-slack/functions/index.js#L38

sean-nicholas:
fixes #100


nicolasgarnier:
Thanks @sean-nicholas !

PR #108, opened 2017-04-05
Switching to using an Express Router

nicolasgarnier:
... rather than an Express app.

That's in order to be less confusing since we are really only routing a sub-path and there is already an express app that is ran somewhere by Cloud Functions.


nicolasgarnier:
cc @katowulf :)

katowulf:
Nice! 🚀🚀🚀

PR #109, opened 2017-04-05
Fixes 'bigquery.dataset' is not a function error

stanvanheumen:
Fixes 'bigquery.dataset' is not a function error after deploying and running the function.


stanvanheumen:
I signed it!

nicolasgarnier:
looks good thanks!

PR #112, opened 2017-04-06
Replace source with event parameter

ahsankhatri:
Replace `paths` with event's parameter because while adding node from firebase console, it's returning absolute url of path and using parameters are the best way to handle such situations.


nicolasgarnier:
Thanks @ahsankhatri! looks good.

PR #116, opened 2017-04-10
ref(...).then is not a function *Fixed*

ahsankhatri:


ahsankhatri:
I signed it!

nicolasgarnier:
Thanks @ahsankhatri !

PR #129, opened 2017-04-26
Fix Typos

jenperson:
There were a few typos in the `quickstarts thumbnails README`.

In addition, I replaced the back ticks with quotes in the `filePath.replace` function because there is no string interpolation so they aren't needed.


jenperson:
I signed it!

jamesdaniels:
Thx Jen :+1:

PR #130, opened 2017-04-27
fix package.json

Ulydev:
Updated some package.json that did not have the correct package name


Ulydev:
Signed it!

nicolasgarnier:
Thanks so much @Ulydev !

PR #131, opened 2017-04-27
Add / to root request. Temp fix #101

### LINKED ISSUE #101: Unable to route the to function when express is used
### How to reproduce these conditions

Firebase deploy the function below:

```js'use strict';

const functions = require('firebase-functions');
const admin = require('firebase-admin');

admin.initializeApp(functions.config().firebase);

const express = require('express');
const cors = require('cors');

const app = express();

app.use(cors({origin: true}));
app.get('*', (req, res) => {
  return res
    .type('application/manifest+json')
    .status(200)
    .send(JSON.stringify({
      message: 'OK',
    }, null, 2));
});

exports.testCloudFunction = functions.https.onRequest(app);

```

### Debug output

* Firebase logs: `Function execution took 235 ms, finished with status code: 404`
* Result on browser when visiting the endpoint: `Cannot GET null`

### Expected behavior

Route to the Firebase's Cloud Function and returns `{ message: 'OK' }` as a result.

### Actual behavior

`Cannot GET null` is display when visiting the endpoint and **a trailing slash is needed to run it properly**.

cdock1029:


cdock1029:
Until https://github.com/firebase/firebase-functions/issues/27 can be handled within framework,
this provides solution for people referencing the examples.

nicolasgarnier:
Sounds good. Thanks @cdock1029 !

PR #132, opened 2017-04-28
Update index.js

jvosloo:
"parent" is now a property and no longer a method
Fixed the bounds so that the max chat messages do not exceed the MAX_LOG_COUNT value


jvosloo:
I signed it!

nicolasgarnier:
Thanks @jvosloo !

jvosloo:
You're welcome :)

PR #133, opened 2017-04-28
Added URL Shortener sample

Ulydev:
Hello,
Here's a sample cloud function I'm using in my project. Thought I'd share it 😃 

It replaces new links with shorter URLs generated with the Google URL Shortener API.


nicolasgarnier:
Thanks for the sample @Ulydev , Sorry for not reviewing it right now as I'm cranking on some stuff for Google IO. I'll review as soon as I can find some time :)

url-shortener/functions/index.js
```diff
@@ -0,0 +1,60 @@
...
+    },
+    json: true,
+    resolveWithFullResponse: true
+  }
+}
+
+function createShortenerPromise(snapshot) {
+  const key = snapshot.key;
+  const originalUrl = snapshot.val();
+  return request(createShortenerRequest(originalUrl)).then(
```
nicolasgarnier: I kind of prefer it this way, i.e. make the Promise return the shortURL then use it inside another `then()`.

```js
return request(createShortenerRequest(originalUrl)).then(response => {
  if (response.statusCode === 200) {
    return response.body.id;
  }
  throw response.body;
}).then(shortUrl => {
  return admin.database().ref(`/links/${key}`).set({
      original: originalUrl,
      short: shortUrl
  });
});

Ulydev:
There you go!

PR #138, opened 2017-05-04
Updating firebase admin.initializeApp to use firebase-functions .config().firebase

ankemp:
This removes the steps the engineer needs to take to create a service account as it uses the firebase-functions` .config().firebase` param to initialize the firebase-admin app.


ankemp:
Dear bot,

I signed it!

With Love.

ankemp:
This would actually require firebase-functions to update to be able to use a custom credential as it currently only uses `admin.credential.applicationDefault()` which doesn't have access to the createCustomToken() method.

Closing this PR.

PR #142, opened 2017-05-10
Fixing a couple typos.

egilmorez:
Just fixing a couple of typos, per a bug from editor.

### DIFF
--- a/coupon-on-purchase/functions/index.js
+++ b/coupon-on-purchase/functions/index.js
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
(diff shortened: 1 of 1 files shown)


coupon-on-purchase/functions/index.js
```diff
@@ -13,15 +13,15 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-'use strict';
+'use strict';licenses
```
nicolasgarnier: That one must be a mistake? :)

nicolasgarnier: Can you remove it? :D

PR #146, opened 2017-05-12
Added developer motivator sample

tidoemanuele:
a simple use case on how to use cloud functions to know each time our app is opened for the first time or is removed from the user device


tidoemanuele:
I signed it!

PR #151, opened 2017-05-18
Add paypal-rest-sdk sample

Kevded:
Use Paypal-rest-sdk on firebase functions
[paypal-rest-sdks](https://developer.paypal.com/docs/api/rest-sdks/)


paypal/functions/index.js
```diff
@@ -0,0 +1,108 @@
+var functions = require('firebase-functions');
```
nicolasgarnier: let's make it all ES2015 style code like the other samples. Can you replace the `var` with `const` ?

paypal/functions/index.js
```diff
@@ -0,0 +1,108 @@
...
+            },
+            // This is the payment transaction description. Maximum length: 127
+            description: 'Cloud Functions Firebase paypal-rest-sdk' // req.body.id
+            // reference_id string .Optional. The merchant-provided ID for the purchase unit. Maximum length: 256.
+            // reference_id: ''
+        }]
+    });
+
+    // Initialize the payment and redirect the user
+    paypal.payment.create(payReq, function (error, payment) {
```
nicolasgarnier: You can use an arrow function here like:

```js
paypal.payment.create(payReq, (error, payment) => {
```

paypal/functions/index.js
```diff
@@ -0,0 +1,108 @@
...
+        var links = {};
+
+        if (error) {
+            console.error(JSON.stringify(error));
```
nicolasgarnier: just:

```
console.error(error);
```

paypal/functions/index.js
```diff
@@ -0,0 +1,108 @@
...
+            res.end();
+        } else {
+            // Capture HATEOAS links
+            payment.links.forEach(function (linkObj) {
```
nicolasgarnier: Use an arrow function here:

```js
payment.links.forEach(linkObj => {
```

paypal/functions/index.js
```diff
@@ -0,0 +1,108 @@
...
+            });
+
+            // If redirect url present, redirect user
+            if (links.hasOwnProperty('approval_url')) {
+                // REDIRECT USER TO links['approval_url'].href
+                // console.log(links['approval_url'])
+                res.set('Content-Type', 'application/json');
+                res.send({
+                    approval_url: links.approval_url.href
+                });
```
nicolasgarnier: I think you can replace the 4 lines above by:

```js
res.json({approval_url: links.approval_url.href});
```

paypal/functions/index.js
```diff
@@ -0,0 +1,108 @@
...
+            } else {
+                console.error('no redirect URI present');
+            }
+            res.status('500').end();
```
nicolasgarnier: That should go inside the `else` right ?

paypal/functions/index.js
```diff
@@ -0,0 +1,108 @@
...
+});
+
+// Complete the payment
+exports.process = functions.https.onRequest((req, res) => {
+    var paymentId = req.query.paymentId;
+    var payerId = {
+        payer_id: req.query.PayerID
+    };
+
+    paypal.payment.execute(paymentId, payerId, function (error, payment) {
```
nicolasgarnier: Use an arrow function here :)

paypal/functions/index.js
```diff
@@ -0,0 +1,108 @@
...
+        if (error) {
+            console.error(JSON.stringify(error));
```
nicolasgarnier: just `console.error(error);`

paypal/functions/index.js
```diff
@@ -0,0 +1,108 @@
...
+            res.redirect('http://localhost:4200/error');
```
nicolasgarnier: Change that URL to a production URL ?

paypal/functions/index.js
```diff
@@ -0,0 +1,108 @@
...
+                    'paid': true,
+                    'description': description,
+                    'date': Date.now()
+                }).then(r => console.info('promise: ', r));
+                // replace url
+                res.redirect('http://localhost:4200/success');
+            } else {
+                console.warn('payment.state: not approved ?');
+                // replace debug url
+                res.redirect('https://console.firebase.google.com/project/<project-id>/functions/logs?search=&severity=DEBUG');
```
nicolasgarnier: You can replace the project ID automatically here as shown above.

paypal/functions/index.js
```diff
@@ -0,0 +1,108 @@
...
+            } else {
+                console.error('no redirect URI present');
+            }
+            res.status('500').end();
+        }
+    });
+});
+
+// Complete the payment
+exports.process = functions.https.onRequest((req, res) => {
```
nicolasgarnier: Could you add  a little bit more details in the comment here to explain a bit more what this function does?

paypal/functions/index.js
nicolasgarnier: Could we also do an HTTP Redirect (like a 302) ?

nicolasgarnier:
ping :)

Kevded:
I work on it. It just misses the comments. Sorry I was taken because of school.

Deddy SAINT-VAL



De : Nicolas Garnier
Envoyé : jeudi 29 juin 14:24
Objet : Re: [firebase/functions-samples] Add paypal-rest-sdk sample (#151)
À : firebase/functions-samples
Cc : kevded, Author


ping :)

nicolasgarnier:
Thanks @Kevded  I've updated the README with more detailed step-by-step instructions.

Do you think it would be possible for you to also commit a small web UI like the one you showed in the animated GIF? this would make the sample so much more complete! PS if you do don't forget about this feature: https://firebase.google.com/docs/hosting/functions this will allow you to have the Paypal Cloud Function on the same domain as the web UI.

Kevded:
I tried to pass the uid value via the optional "custom" field in the "transaction" object as in paypal [https://developer.paypal.com/docs/api/payments/#definition-transaction](https://developer.paypal.com/docs/api/payments/#definition-transaction), but this does not work, so I have passed through the "description" field.

nicolasgarnier:
I added a few style fixes and we should be all set :)

PR #153, opened 2017-05-19
Typo on comment

JoaquimLey:
- In the current use-case, the user subscribed to the newsletter instead of unsubscribed.


nicolasgarnier:
Thanks @JoaquimLey !

PR #161, opened 2017-05-24
Catch block

jenperson:
Added catch block to email confirmation sample.

### DIFF
--- a/email-confirmation/functions/index.js
+++ b/email-confirmation/functions/index.js
@@ -46,7 +46,7 @@ exports.sendEmailConfirmation = functions.database.ref('/users/{uid}').onWrite(e
     return mailTransport.sendMail(mailOptions).then(() => {
       console.log('New subscription confirmation email sent to:', val.email);
     }).catch(error => {
-      console.log(error);  
+      console.error('There was an error while sending the email:', error);  
     });
   }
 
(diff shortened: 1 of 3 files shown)


inlined:
Everything LGTM except the catch block. Errors should probably be logged to console.error so they can be filtered for more easily. This type of catch also handles the error, making the function evaluate as having been successful. If this is intentional a comment would be good. Otherwise if you can just return the error and the Firebase Functions SDK will add a failure handler for you. If you want to handle an error but still fail (e.g. add a message) you can have `.catch` throw an error or return `Promise.reject(myValue)`

email-confirmation/functions/index.js
```diff
@@ -45,6 +45,8 @@ exports.sendEmailConfirmation = functions.database.ref('/users/{uid}').onWrite(e
     mailOptions.text = 'Thanks you for subscribing to our newsletter. You will receive our next weekly newsletter.';
     return mailTransport.sendMail(mailOptions).then(() => {
       console.log('New subscription confirmation email sent to:', val.email);
+    }).catch(error => {
+      console.log(error);
```
nicolasgarnier: just do `console.error('There was an error while sending the email:', error)` here

email-confirmation/functions/index.js
```diff
@@ -53,5 +55,7 @@ exports.sendEmailConfirmation = functions.database.ref('/users/{uid}').onWrite(e
   mailOptions.text = 'I hereby confirm that I will stop sending you the newsletter.';
   return mailTransport.sendMail(mailOptions).then(() => {
     console.log('New unsubscription confirmation email sent to:', val.email);
+  }).catch(error => {
+    console.log(error);
```
nicolasgarnier: just do `console.error('There was an error while sending the email:', error)` here :)

nicolasgarnier:
Actually just added the fix myself to your branch since this was a small change :)

PR #162, opened 2017-05-24
user is not defined

carlosazaustre:
user is in `event.params.userId`, if not is not defined


nicolasgarnier:
Thanks for the fix @carlosazaustre !

PR #169, opened 2017-06-07
Stripe - Update Readme to Fix project configuration

nhterry:
There was no mention to make sure the developer had Firebase CLI Tools installed, and the command to configure the sample to use the project was incorrect.
`firebase --use add` --> `firebase use --add`


nhterry:
I signed it!

nicolasgarnier:
Thanks @nhterry !

PR #176, opened 2017-06-16
Need to create temp directory where storage file will be downloaded

ravishan110:
We need to create the temp directory where storage file will be downloaded and stored, before downloading it, otherwise error : `ENOENT: no such file or directory,` will be thrown. 

To create temp directory, `mkdirp-promise` is used. After successful creation of the directory, file is downloaded.

### DIFF
--- a/exif-images/functions/index.js
+++ b/exif-images/functions/index.js
@@ -17,6 +17,7 @@
 
 const functions = require('firebase-functions');
 const mkdirp = require('mkdirp-promise');
+const fs = require('fs');
 
 const admin = require('firebase-admin');
 admin.initializeApp(functions.config().firebase);
@@ -64,6 +65,12 @@ exports.metadata = functions.storage.object().onChange(event => {
       return admin.database().ref(makeKeyFirebaseCompatible(filePath)).set(metadata).then(() => {
         console.log('Wrote to:', filePath, 'data:', metadata);
       });
+    });
+  }).then(()=>{
+    //cleanup temp directory after metadata is extracted
+    //Remove the file from temp directory
+    return fs.unlink(tempLocalFile,()=>{
+      console.log("cleanup successful!");
     });
   });
 });
--- a/exif-images/functions/index.js
+++ b/exif-images/functions/index.js
@@ -16,6 +16,8 @@
 'use strict';
 
 const functions = require('firebase-functions');
+const mkdirp = require('mkdirp-promise');
+
 const admin = require('firebase-admin');
 admin.initializeApp(functions.config().firebase);
 const gcs = require('@google-cloud/storage')();
@@ -30,6 +32,9 @@ exports.metadata = functions.storage.object().onChange(event => {
   const object = event.data;
   const filePath = object.name;
   const fileName = filePath.split('/').pop();
+  const fileDir = filePathSplit.join('/') + (filePathSplit.length > 0 ? '/' : '');
+
+  const tempLocalDir = `${LOCAL_TMP_FOLDER}${fileDir}`;
   const tempLocalFile = `${LOCAL_TMP_FOLDER}${fileName}`;
 
   // Exit if this is triggered on a file that is not an image.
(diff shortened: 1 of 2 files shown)


ravishan110:
I signed it, please review my pull request.

exif-images/functions/index.js
```diff
@@ -16,6 +16,8 @@
 'use strict';
 
 const functions = require('firebase-functions');
+const mkdirp = require('mkdirp-promise');
```
ravishan110: required for creating temp directory, before downloading file from bucket

exif-images/functions/index.js
```diff
@@ -30,6 +32,9 @@ exports.metadata = functions.storage.object().onChange(event => {
   const object = event.data;
   const filePath = object.name;
   const fileName = filePath.split('/').pop();
+  const fileDir = filePathSplit.join('/') + (filePathSplit.length > 0 ? '/' : '');
+
+  const tempLocalDir = `${LOCAL_TMP_FOLDER}${fileDir}`;
```
ravishan110: temporary directory location

exif-images/functions/index.js
```diff
@@ -44,10 +49,13 @@ exports.metadata = functions.storage.object().onChange(event => {
     return;
   }
 
-  // Download file from bucket.
   const bucket = gcs.bucket(object.bucket);
-  return bucket.file(filePath).download({
-    destination: tempLocalFile
+  //create the temp directory where storage file will be downloaded
+  return mkdirp(tempLocalDir).then(()=>{
```
ravishan110: first create the temp directory, and download the file after successful creation of the directory

inlined:
Can you please update the example to clean up the tempdir after the function is finished? The size of /tmp counts against the Cloud Functions storage quota.

puf:
It seems that last request now made it onto Stack Overflow: https://stackoverflow.com/questions/44653533/cleanup-temp-directory-firebase-cloud-functions/44654557?noredirect=1#comment76295491_44654557

ravishan110:
@puf Yeah, i had some options to remove file from temp directory, but i wanted to be sure before doing any changes. that's why i asked a question on Stack Overflow. Though this question was on my mind before this pull request was made, when i was writing functions for my webapp. 

Can child-process-promise be used to remove file like this:
```
const exec = require('child-process-promise').exec;
exec(`rm -rf {$tempLocalFile}`).then(()=>{
...
})
```
Though the above function ran without any error, but there was no way i could see inside the temp directory to check if the file was actually removed or not.

One another option was `fs.unlink()` instead of `fs.rmdir()` , as some other function might be using `temp` directory, and it might not be good to remove the directory itself.

Just want to be sure about my method before commiting any changes.
Any help or guidance is greatly appreciated. Thanks!

ravishan110:
@inlined I have added cleanup code. Please review my changes.

exif-images/functions/index.js
```diff
@@ -57,6 +66,12 @@ exports.metadata = functions.storage.object().onChange(event => {
         console.log('Wrote to:', filePath, 'data:', metadata);
       });
     });
+  }).then(()=>{
+    //cleanup temp directory after metadata is extracted
+    //Remove the file from temp directory
+    return fs.unlink(tempLocalFile,()=>{
```
inlined: nit that we should also unlink the directory

exif-images/functions/index.js
```diff
@@ -44,10 +50,13 @@ exports.metadata = functions.storage.object().onChange(event => {
     return;
   }
 
-  // Download file from bucket.
   const bucket = gcs.bucket(object.bucket);
-  return bucket.file(filePath).download({
-    destination: tempLocalFile
+  //create the temp directory where storage file will be downloaded
+  return mkdirp(tempLocalDir).then(()=>{
```
inlined: nit: spaces around =>

exif-images/functions/index.js
inlined: also, nit for spaces around => again

ravishan110:
@inlined  I checked the code with your fix, It works completely fine. But there still is one problem with this (i think!). 
All the files are downloaded directly into the temp folder. what will happen if two files (which are stored at different location in firebase storage) have the same name. If these files are uploaded at nearly the same time, In that case the second file will overwrite the first file during the operation, which in turn will give the wrong result (both will have same metadata). Though the probability of this happening is very less, i guess. but i just have this concern.
In that case, creating a directory inside temp folder and downloading the uploaded file in that folder would be better choice. And, later we need to unlink that directory as well after unlinking the file(as you suggested).

inlined:
Ah, great catch! This isn't a problem in practice since a Cloud Function only handles a single request at a time, but I don't like people to write code that depends on this. We could fix this by creating a random file in each execution rather than using the base name. This seems at least as good as creating a random directory with a known file.

ravishan110:
@inlined i guess i handled everything now. Please have a look at it.
Removed `mkdir`, as it was no longer necessary. Thus, we only need to `unlink` the temporary file.

inlined:
Love it!

PR #180, opened 2017-06-22
Use Shrap and streams to create thumbnail

ovaris:
Currently there is a problem using Imagemagick and/or Node process spawn in Firebase Functions environment, it leaks memory heavily and you will get `Function killed. Error: memory limit exceeded` most probably.

This PR uses Sharp & Node.js Streams to do the same but much much more efficiently. 

Note: this function is untested, but I run similar function in own project and it performs beautifully.


nicolasgarnier:
Hey @ovaris,

That's a nice sample! However the quickstart sample here is used in the docs and we do mention having imagemagick pre-installed on all instances etc (also I'm fixing this samlpe right now to cleanup the file and use `path` to manipulate path)\

Instead could you create a new sample with what you made and call it `thumbnail-sharp` on the top level folder? And add an entry in the main README for it.

cheers!

prescottprue:
@ovaris When trying to use this with rectangular images, I had the tops and bottoms be cut off (which does not happen with the imagemagick example).

ovaris:
I signed it!

ovaris:
@prescottprue Thanks for the tip to use `max()`! I've only used width (and not passing height) which will retain the original ratio, will fix.

ovaris:
Thanks for the input @nicolasgarnier, I did like you suggested, moved this sample from quickstarts to parent folder, cheers.

nicolasgarnier:
Thanks @ovaris!

sarovin:
hi,
thanks for this example @ovaris ...but i can't make it work.
I think that the problem is this => https://github.com/lovell/sharp/issues/462

PR #185, opened 2017-06-30
fixed file is not a function error

th3brink:
Line 77 was trying to pass filePath into file, but file was already instantiated with filePath.


nicolasgarnier:
thanks @th3brink !

PR #201, opened 2017-08-04
sample typescript setup

ultrasaurus:
Simple example with TypeScript setup.  
README has step-by-step instructions


typescript-getting-started/README.md
```diff
@@ -0,0 +1,143 @@
...
+yarn install
+```
+7. create a `tsconfig.json` file:
+```
+{
+  "compilerOptions": {
+    "lib": ["es6", "es2015.promise"],
+    "module": "commonjs",
+    "noImplicitAny": false,
+    "outDir": "",
```
inlined: A more typical setup is probably to use a "lib" dir and then put that directory in gitignore.

typescript-getting-started/firebase.json
```diff
@@ -0,0 +1 @@
+{}
```
inlined: Is this file here so one can use inline commands? e.g. `firebase deploy --project foo`?

typescript-getting-started/README.md
```diff
@@ -0,0 +1,143 @@
+## Cloud Functions for Firebase: getting started with TypeScript
+
+This is an example Firebase project for using TypeScript with
+[Cloud Functions for Firebase](https://firebase.google.com/products/functions)
+
+Pre-requistes:
```
inlined: This is a little odd to have both a fully functional template project and then instructions that don't use them. Maybe the steps for manual setup can be in DIY.md and linked from main README.md?

typescript-getting-started/README.md
```diff
@@ -0,0 +1,143 @@
...
+npm install -g firebase-tools
+```
+
+Also, I highly recommend using [yarn](https://yarnpkg.com),
+which locks the versions of your npm modules.  I also like to use
+[homebrew](https://brew.sh/) to install dependencies on the Mac.  These aren't
+required for Cloud Functions or Firebase, but are used in the instructions.
+
+```
+brew install yarn
```
inlined: `npm install -g yarn` is more likely to work on the developer's machine. Then the dev wouldn't have to go through the work of a heavy install to try out typescript either.

typescript-getting-started/README.md
```diff
@@ -0,0 +1,143 @@
...
+? Which Firebase CLI features do you want to setup for this folder? Press Space to select features, then Enter
+ to confirm your choices. Functions: Configure and deploy Cloud Functions
+
+=== Project Setup
+
+First, let's associate this project directory with a Firebase project.
+You can create multiple project aliases by running firebase use --add,
+but for now we'll just set up a default project.
+
+i  .firebaserc already has a default project, skipping
```
inlined: I'm a bit confused why this would happen if they've just created their project.

typescript-getting-started/README.md
```diff
@@ -0,0 +1,143 @@
...
+
+=== Functions Setup
+
+A functions directory will be created in your project with a Node.js
+package pre-configured. Functions can be deployed with firebase deploy.
+
+✔  Wrote functions/package.json
+✔  Wrote functions/index.js
+? Do you want to install dependencies with npm now? No
```
inlined: Why not?

typescript-getting-started/README.md
```diff
@@ -0,0 +1,143 @@
...
+```
+5. add the following dev dependencies and scripts to functions/package.json
+```
+  "devDependencies": {
+    "typescript": "^2.3.2"
+  },
+  "scripts": {
+    "build": "tsc",
+    "watch": "tsc --watch",
+    "deploy": "tsc && firebase deploy --only functions"
```
inlined: <3 adding scripts!
1. @mbleigh may know better than me what the search order for NPM scripts is. I've been using "./node_modules/.bin/tsc" to get IDEs to stop using a global version of `tsc`.
2. To help nip user errors in the bud, I'd also love to have a lint rule that uses a `tslint.json` with sane defaults (like not handling a Promise)

typescript-getting-started/README.md
```diff
@@ -0,0 +1,143 @@
...
+      "node_modules/@types"
+    ]
+  },
+  "include": [
+    "src/**/*.ts",
+    "spec/**/*.ts"
+  ]
+}
+```
+8. I like to exclude node_modules and compiled js files from git, so
```
inlined: I like this too! I'm not sure if it's pedantic to also suggest using a `.npmignore` file that excludes IDE/linter files + TypeScript source.

typescript-getting-started/README.md
```diff
@@ -0,0 +1,143 @@
...
+I add this to a root level `.gitignore` file
+```
+node_modules/
+jspm_packages/
```
inlined: I didn't know about jspm. It looks like this is a browser version of NPM? Will the dev ever need this?

typescript-getting-started/README.md
```diff
@@ -0,0 +1,143 @@
...
+
+## ignore generated JavaScript files
+functions/**/*.js
+functions/**/*.js.map
+
+```
+9. in `index.ts` remove comments around the sample function (and
+add a couple of newlines to make the output easier to see):
+```
+exports.helloWorld = functions.https.onRequest((request, response) => {
```
inlined: Directly manipulating the exports object is an anti-pattern in TypeScript because it's not ES6 module compatible. I didn't even think this would compile. The normal syntax (which also works in our version of Node IIRC) is

```
export let helloWorld = functions.https.onRequest((request, response) => {
```

Alternatively you could have broken that into two lines:
```
let helloWorld = functions...
export {helloWorld};
```

typescript-getting-started/README.md
inlined: Depending on how heavy handed you wanted to be with best practices, you could even add a spec directory and a test script. Honestly I haven't found a code layout I love for tests & prod with TypeScript, so it might be worthwhile.

typescript-getting-started/README.md
```diff
@@ -0,0 +1,143 @@
...
+You will see a bunch of output and at the end it will show you the URL for your deployed function.
+
+10. test YOUR function with curl. For mine I can do this:
+```
+curl https://us-central1-functions-typescript.cloudfunctions.net/helloWorld
+```
+
+Now start developing in TypeScript!
+
+Initial steps were based on [nice article](https://medium.com/@wcandillon/writing-cloud-functions-with-typescript-61d86e282752) by [@wcandillon](https://twitter.com/wcandillon) which
```
inlined: Looks like this is missing its closing thoughts.

typescript-getting-started/functions/tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
+  "compilerOptions": {
+    "lib": ["es6", "es2015.promise"],
+    "module": "commonjs",
+    "noImplicitAny": false,
+    "outDir": "",
+    "sourceMap": true,
+    "target": "es5",
```
inlined: target es6 to improve the readability and presumedly speed of the transpiled source.

typescript-getting-started/functions/tsconfig.json
```diff
@@ -0,0 +1,17 @@
...
+    "typeRoots": [
```
inlined: This is now a default and can be omitted.

typescript-getting-started/functions/tsconfig.json
```diff
@@ -0,0 +1,17 @@
...
+      "node_modules/@types"
+    ]
+  },
+  "include": [
```
inlined: CC @mbleigh. Have you ever settled on a tsconfig file that you love? I forget whether this just dumps all files into the top-level directory or embeds everything as "$OUTDIR/[src|spec]/**/*.ts". I think our big options are:

1. Compile everything in-place so a .ts, .js, .js.map file exists at the same level. Pro: easy to set up, Con: hard to get clean builds, not all tools support file globs (e.g. gitignore can't specifically ignore *.js in nested directories of a single directory, so this breaks hybrid codebases)
2. Compile src & spec into a lib directory. Pros: easy. Cons: lib/src/*.js is confusing as heck
3. Use a tsconfig.json and a tsconfig.prod.json that inherits it. `lib` only contains compiled versions of `src` while `test` contains compiled versions of `src` and `spec`. Pros: production code doesn't have tests. Cons: more complex. Breaks anyone who assumes `tsc && firebase deploy` is the right command.

inlined:
Very excited for this. Would love an intro in the README.md that explains why this is so awesome as well.

typescript-getting-started/README.md
ultrasaurus: good idea

typescript-getting-started/README.md
ultrasaurus: I'll just skip the yarn install, there are details at the yarn link

typescript-getting-started/README.md
ultrasaurus: I don't know, I just copied the output that I got

typescript-getting-started/README.md
ultrasaurus: This is actually the first time I ever set up TypeScript myself... not sure exactly how I would use a lib directory...

typescript-getting-started/README.md
ultrasaurus: I read up on `. npmignore` -- looks like it is just for packages (where I thought this was an app that uses packages).   Or... how does it apply to this context?

typescript-getting-started/README.md
ultrasaurus: Good point, that crept in from some other experiment.  Will remove.

typescript-getting-started/README.md
ultrasaurus: Oops. Removed "which"

typescript-getting-started/firebase.json
ultrasaurus: `firebase init` creates this file

typescript-getting-started/functions/tsconfig.json
ultrasaurus: good idea

typescript-getting-started/functions/tsconfig.json
ultrasaurus: good to know!

typescript-getting-started/README.md
ultrasaurus: ah, when I switched tsconfig to target `es6` then I had to fix this (and also, hadn't noticed that I should have switched the `require` to `import` so that was a good catch!

typescript-getting-started/functions/tsconfig.json
ultrasaurus: yeah, I don't love dumping all the js files into the top level, but didn't know how to specify where `index.js` is if it is not at the root of the `functions` directory.  

The way this is set up now, all the `.ts` files are in `src` sub-directory, so at least that is something.

ultrasaurus:
@inlined I addressed all of the feedback or added responses on github where I wasn't sure.  I took a stab at explaining why TypeScript is awesome... feel free to suggest edits to that part (or anything else, of course)

Thanks!

ultrasaurus:
@mbleigh I usually use `lib` for actually libraries that I include in my source, but if its a convention you like, happy to make the change.  The other name I've seen people use is `build` which I also don't love, but like a little better.

I like the idea of putting the compiled js files in their own sub-directory.  

Thoughts on naming?

ultrasaurus:
@inlined I talked to @joelpoloney and he had similar reaction to `lib` and had also seen `build` used elsewhere.  I've updated the code to use `build` as the sub-directory for compiled JS.

Can you take a look when you get a chance?

CodingDoug:
I like also "build" for build artifacts - comfortable for Android/Gradle users.

typescript-getting-started/README.md
```diff
@@ -0,0 +1,69 @@
...
+that your deployed Cloud Functions are using the exact same versions of
+npm modules that you are running locally. The specific version of each
+dependency is saved in [yarn.lock](functions/yarn.lock)
+
+
+This example has the default sample function. To use this example as a
+starter project:
+
+1. `npm install -g firebase-tools`
+2. `git clone https://github.com/firebase/functions-samples.git`
```
inlined: oof. I'm not sure if this is clearer or not, but I was able to tinker and find this:

```bash
curl https://codeload.github.com/firebase/functions-samples/tar.gz/master |  \
  tar -xz --strip=2 functions-samples-*/typescript-getting-started
```

This has the benefit of not creating a git repo that may not be compatible with any existing repros and gets rid of the need for the `rm -rf` & `cp -r` lines below. Don't really have an opinion which is better; just thought I'd give you an FYI.

typescript-getting-started/firebase.json
inlined: Sure; but if `firebase init` is part of the setup instructions then we don't need this. Not a huge deal either way.

typescript-getting-started/HOW-TO.md
```diff
@@ -0,0 +1,140 @@
...
+
+```
+node_modules/
+
+## ignore generated JavaScript files
+functions/**/*.js
+functions/**/*.js.map
+```
+
+9. in `index.ts` update the syntax to TypeScript (`require` -> `import` and
```
inlined: This is really good. It might be worth putting in the main page. WDYT about some section like "what's different?". The other thing we should probably mention is that people will also want to replace `npm install --save foo` with `npm install --save foo @types/foo` to get TypeScript type information where possible.

typescript-getting-started/functions/src/index.ts
```diff
@@ -0,0 +1,12 @@
+import * as functions from 'firebase-functions'
+
+// if you need to use the Firebase Admin SDK, uncomment the following:
+// import * as admin from 'firebase-admin'
+
+
+// // Create and Deploy Your First Cloud Functions
```
inlined: Double comments are weird here. I'd probably either leave the whole function commented or remove the double comments & empty comment line.

typescript-getting-started/functions/package.json
```diff
@@ -0,0 +1,18 @@
...
+  },
+  "devDependencies": {
+    "typescript": "^2.3.2"
+  },
+  "scripts": {
+    "build": "tsc",
+    "watch": "tsc --watch",
+    "deploy": "tsc && firebase deploy --only functions"
+  },
+  "main": "build/index.js",
```
inlined: I'm surprised this works; it seems like a TypeScript bug. I think that if you had `src/index.ts` and `spec/index.spec.ts` you'll end up getting `build/src/index.js` build/spec/index.js`

ultrasaurus: well, this example doesn't have tests yet, I'd be inclined not to put output for running tests in the build directory -- for example: could be spec/src and spec/build.  In any case, it works as is, and we can circle back and add tests later

typescript-getting-started/README.md
ultrasaurus: Nice suggestion.  I agree it is more verbose as is, but I think it requires less knowledge of bash, so will leave it.

typescript-getting-started/README.md
```diff
@@ -0,0 +1,69 @@
...
+This project was set up using [yarn](https://yarnpkg.com) which will ensure
+that your deployed Cloud Functions are using the exact same versions of
+npm modules that you are running locally. The specific version of each
+dependency is saved in [yarn.lock](functions/yarn.lock)
+
+
+This example has the default sample function. To use this example as a
+starter project:
+
+1. `npm install -g firebase-tools`
```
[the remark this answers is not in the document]
ultrasaurus: I looked into this.  You are correct.  I will see if we can get this written up as something we track in the public bug list, so folks are aware.

typescript-getting-started/HOW-TO.md
```diff
@@ -0,0 +1,140 @@
...
+9. in `functions` directory, use the npm build script created
+above to build TypeScript files and deploy:
+```
+npm run deploy
+```
+You will see a bunch of output and at the end it will show you the URL for your deployed function.
+
+10. test YOUR function with curl. For mine I can do this:
+```
+curl https://us-central1-functions-typescript.cloudfunctions.net/helloWorld
```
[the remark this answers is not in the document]
ultrasaurus: nice -- added!

typescript-getting-started/HOW-TO.md
ultrasaurus: updated README with a section on this

typescript-getting-started/functions/package.json
inlined: The concern is whether the production code would break simply by adding a test file.

PR #206, opened 2017-08-08
Automatically discover the data to be deleted.

eobrain:
The user-data-cleanup example is now a library to help Firebase developers increase privacy of their end-users. The library uses Cloud Functions to automatically find and delete sensitive user data at account deletion, based on Security Rules.


eobrain:
Those commits have the wrong or misconfigured e-mail address associated with them.  The are in fact all by Github user horseno who is an intern at Google.

user-data-cleanup/functions/access.js
```diff
@@ -0,0 +1,139 @@
...
+
+const exp = require('./expression');
+
+/**
+ * Access Class, used to represent the access status of a write rule or node
+ * @param status access status, could be NO_ACCESS/SINGLE_ACCESS/MULT_ACCESS
+ * @param list variable list, should be empty list if status is NO/SINGLE,
+ * else should be the list of literal in the conjunction of corresponding exp
+ */
+function Access(status, list) {
```
nicolasgarnier: Just FYI, since this runs on Functions and Functions run on Node 6 we could have used the ES2015 `class` notation here. I'd say this is fine though but if you want/have time to re-write in class notation it would make it a bit easier to read :) (I know there isn't that much time left so it's fine to leave as-is also :))

nicolasgarnier: So just let me know if you want to move forward here or leave as is!

user-data-cleanup/README.md
```diff
@@ -22,7 +36,8 @@ When a user signs-in we write some sample personal data of the following form:
         $uid : "Some user data"
```
nicolasgarnier: lmk if you prefer taking care of this in a subsequent PR.

nicolasgarnier:
SG then, the rest looks good :) Let's merge!

PR #210, opened 2017-08-14
user-data-cleanup: Support condition and data references for auto extraction of user-data locations from RTDB rules

eobrain:


nicolasgarnier:
LGTM!

Issue #172, opened 2017-06-08
Stripe - Upgrading to Firebase 4.1, Stripe v3


nhterry:
I'm having a hard time figuring out how to upgrade this example to the latest and greatest Firebase and Stripe JS SDKs.

https://github.com/firebase/functions-samples/tree/master/stripe

**How to reproduce these conditions**
In https://github.com/firebase/functions-samples/blob/master/stripe/public/index.html
Add: `<script src="https://js.stripe.com/v3/"></script>` after the v2 script.
So:
```
  <script src="https://js.stripe.com/v2/"></script>
  <script src="https://js.stripe.com/v3/"></script>
```

Replace:
```
    <script src="/__/firebase/3.9.0/firebase-app.js"></script>
    <script src="/__/firebase/3.9.0/firebase-auth.js"></script>
    <script src="/__/firebase/3.9.0/firebase-database.js"></script>
    <script src="/__/firebase/init.js"></script>
```
with:
```
   <script src="/__/firebase/4.1.1/firebase-app.js"></script>
    <script src="/__/firebase/4.1.1/firebase-auth.js"></script>
    <script src="/__/firebase/4.1.1/firebase-database.js"></script>
    <script src="/__/firebase/init.js"></script>
```

You may also need to switch the Stripe API Key from:
```
      var stripe = Stripe('pk_test_XXXXXXXXXXXXXXXXXXXX');
```
to:
```
      Stripe.setPublishableKey('pk_test_XXXXXXXXXXXXXXXXXXXX');
```

**Sample name or URL where you found the bug**
https://github.com/firebase/functions-samples/tree/master/stripe
```
firebase/functions-samples/stripe
```

**Failing Function code used (including require/import commands at the top)**
It's not really the function code that seems to be failing, but listening to whether or not the customer has a stripe id in the firebase database. I'm able to create the customer/account in Firebase and see them in Stripe, but I am unable to tell if the stripe customer is initialized or to set stripeCustomerInitialized to be true.

```
            firebase.database().ref(`/stripe_customers/${this.currentUser.uid}/customer_id`).on('value', snapshot => {
              this.stripeCustomerInitialized = (snapshot.val() !== null);
            }, () => {
              this.stripeCustomerInitialized = false;
            });
```

When, it listens for this function, even though a customer is created and has values within Firebase, this.stripeCustomerInitialized always sets to be false.

Steps to set up and reproduce

**Sample data pasted or attached as JSON (not an image)**

```
{
[...]

PR #227, opened 2017-08-30
Firebase version updated from 3.9.0 to 4.2.0

NoahWilcox:
Upgrading to the latest Firebase will solve the StripeCustomerInitialized always being set to false for issue #172. Even though the version used in the issue is different from the version in the source code, I think updating to the latest version would be beneficial.


nicolasgarnier:
Thanks @NoahWilcox

Issue #126, opened 2017-04-23
Stripe clean up user bug


SamMatthewsIsACommonName:
The clean up user function is a little at odds with the stripe API and returned an error for me. It should be changed to:

````
// When a user deletes there account, clean up after them
exports.cleanupUser = functions.auth.user().onDelete(event => {
  return admin.database().ref(`/stripe_customers/${event.data.uid}`).once('value').then(snapshot => {
    return snapshot.val();
  }).then(customer => {
// change below, customer.customer_id vs customer
    return stripe.customers.del(customer.customer_id);
  }).then(() => {
    return admin.database().ref(`/stripe_customers/${event.data.uid}`).remove();
  });
});
````
OR change the reference to admin.database().ref(`/stripe_customers/${event.data.uid}/customer_id`)
Thanks

katowulf:
@SamMatthewsIsACommonName can you please follow the issue template and include the necessary details here? Cheers.

SamMatthewsIsACommonName:
Well I don't think it really needs to be replicated much as it seems to be just one of couple of typos (the other one I saw is int he database rules $sourceId and $chargeId are around the wrong way).
I think the author intended to follow the same pattern as when adding a payment source, by getting a read from 

admin.database().ref(`/stripe_customers/${event.params.userId}/customer_id`).once('value')

to get the stripe customer id as a string:

// Add a payment source (card) for a user by writing a stripe payment source token to Realtime database
exports.addPaymentSource = functions.database.ref('/stripe_customers/{userId}/sources/{pushId}/token').onWrite(event => {
  const source = event.data.val();
  if (source === null) return null;
  return admin.database().ref(`/stripe_customers/${event.params.userId}/customer_id`).once('value').then(snapshot => {
    return snapshot.val();
  }).then(customer => {
    return stripe.customers.createSource(customer, {source});
  }).then(response => {
      return event.data.adminRef.parent.set(response);
    }, error => {
      return event.data.adminRef.parent.child('error').set(userFacingMessage(error)).then(() => {
        return reportError(error, {user: user});
      });
  });
});

However I think they forgot to include this path n the delete method, as it's also expecting the string 'customer_id' value as the argument, according to the stripe API (https://stripe.com/docs/api#delete_customer).

// When a user deletes there account, clean up after them
exports.cleanupUser = functions.auth.user().onDelete(event => {
  return admin.database().ref(`/stripe_customers/${event.data.uid}`).once('value').then(snapshot => {
[...]

PR #228, opened 2017-08-30
Fixes the cleanupUser cloud function.

NoahWilcox:
Fixes the cleanupUser cloud function as mentioned in #126.

The error thrown in cloud function logs is as follows:
cbf3abc9c


nicolasgarnier:
Thanks @NoahWilcox !

PR #229, opened 2017-08-31
Remove index.js from top-level .gitignore

eobrain:
Given index.js is a checked-in file in many of the subdirectories, I assume that it should not be in the top-level index.js.

Maybe it was intended to be in the .gitignore files of some subdirectories where the index.js is generated rather than checked-in.


nicolasgarnier:
Looks like this was a mistake :) Thanks for removing!

PR #230, opened 2017-09-01
Remove unneeded "api" segment in path

LloydVincent:
Playing around with this code, it seems ```exports.api``` links the path to "/api" automatically, it's redundant to add it in the get/post path and causes the url to fail.


LloydVincent:
I signed it

nicolasgarnier:
Thanks @LloydVincent that's right, the 'api' prefix was never needed :)

PR #232, opened 2017-09-01
link TypeScript example (grouped with similar sample)

ultrasaurus:
linked TypeScript example with a little description
I also moved the two dev env setup samples to their own section


nicolasgarnier:
Thanks for this @ultrasaurus 

What about adding the following 2 samples to the "Development Environment Samples" section: `isomorphic-react-app` and `template-handlebars` ?

nicolasgarnier:
I added a bunch of categories. Merging!

PR #248, opened 2017-10-06
Removes usage of @google-cloud/firestore

abehaskins:
Removes old snippet which uses `@google-cloud/firebase`

### DIFF
--- a/fulltext-search-firestore/functions/package.json
+++ b/fulltext-search-firestore/functions/package.json
@@ -5,7 +5,6 @@
     "algoliasearch": "^3.24.0",
     "cors": "^2.8.4",
     "express": "^4.16.1",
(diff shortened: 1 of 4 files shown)


nicolasgarnier:
fulltext-search-firestore and quickstart/uppercaser-firestore also import @google-cloud/firebase could you premove it from there too?

Also can you remove the import in package.json ? (in all 3 samples)

abehaskins:
Are you sure? I don't see them importing it, however it is in the package.json and I'll fix that.

nicolasgarnier:
Yup I think it's just in the 3 package.json :)

abehaskins:
Okay, removed!

nicolasgarnier:
Looks good!

PR #251, opened 2017-10-10
Change all `firebase --use add` to `firebase use --add`

whhone:
Change all wrong examples of `firebase --use add` to `firebase use --add`.


nicolasgarnier:
Thanks @whhone !

PR #266, opened 2017-10-31
Update copy for Crashlytics in readme

trekforever:


nicolasgarnier:
thanks @trekforever ! :)

PR #275, opened 2017-11-13
Optimized send email confirmation method

Brianvdb:
- Line 39: Always return value (never return undefined);
- Optimized promise chaining & simplified if checks.


Brianvdb:
I signed it!

nicolasgarnier:
Thanks @Brianvdb ! lgtm

PR #279, opened 2017-11-20
Thumbnail errors

inlined:
To help better diagnose customer issues with the thumbnailer sample, I've added stdout/stderr capturing to the spawn command.

When testing, I noticed that my credentials were left as dirty files. Added them to the gitignore so nobody can accidentally commit a certificate file.


nicolasgarnier:
Looks good thanks @inlined !

PR #295, opened 2017-12-15
Transpile sample: Switch to babel-preset-env

### LINKED ISSUE #294: Switch to babel-preset-env
Ran into a few issues of orders of babel plugins causing things to break with node v6.11.* (needed [for functions runtime](https://cloud.google.com/functions/docs/writing/#the_cloud_functions_runtime)). I have found `babel-preset-env` to be great for this since it automatically detects what you need. This way it is also not locked down to `es2017` specifically.

[This config](https://github.com/firebase/functions-samples/blob/105dfec8a48268bedcae84e86e028df39b160ebc/es2017-transpile/functions/.babelrc#L2) would just need to become something like:

```json
{
  "presets": [
    ["env", {
      "targets": {
        "node": "6.11"
      }
    }]
  ]
}
```

If you are using [rest spread](https://babeljs.io/docs/plugins/transform-object-rest-spread/) it is also necessary to add the plugin directly:

```json
{
  "presets": [
    ["env", {
      "targets": {
        "node": "6.11"
      }
    }]
  ],
  "plugins": [
    "transform-object-rest-spread",
  ]
}

```

nicolasgarnier:
fixes #294

### DIFF
--- a/es2017-transpile/functions/.babelrc
+++ b/es2017-transpile/functions/.babelrc
@@ -1,6 +1,13 @@
 {
-  "presets": ["es2017"],
+  "presets": [
+    ["env", {
+      "targets": {
+        "node": "6.11"
+      }
+    }]
+  ],
   "plugins": [[
-    "transform-runtime", {"polyfill": false, "regenerator": true}
+    "transform-runtime", {"polyfill": false, "regenerator": true},
+    "transform-object-rest-spread"
   ]]
 }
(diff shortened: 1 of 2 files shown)


nicolasgarnier:
@prescottprue PTAL :)

prescottprue:
@nicolasgarnier Looks great to me 💯 , its the same changes I was going to make for a PR

nicolasgarnier:
Thanks @prescottprue !

PR #313, opened 2018-01-20
Fix error on startup

The-Alchemist:
There is no `serviceAccount.json`, so `serve` and `deploy` fail


nicolasgarnier:
Hey @The-Alchemist 
The README of the sample explains where to get the serviceAccounts file for your Firebase project.

It is needed because some auth operations can't be done with the default service credentials. Like creating a Custom Auth token for instance.

Let me know if you were having issues with the setup.

PR #319, opened 2018-01-31
fix tslint path to project config

a-xin:
`-p` (or the long version `--project`) was incorrectly referencing the linting configuration in `tslint.json`. Fixing this to reference the `tsconfig.json` file to enable rules that work with the type checker.

Using the long version `--project` as it is more readable

### DIFF
--- a/typescript-getting-started/functions/package.json
+++ b/typescript-getting-started/functions/package.json
@@ -11,7 +11,7 @@
     "typescript": "^2.6.1"
   },
   "scripts": {
-    "build": "./node_modules/.bin/tslint -p tslint.json && ./node_modules/.bin/tsc",
+    "build": "./node_modules/.bin/tslint --project tsconfig.json && ./node_modules/.bin/tsc",
     "serve": "npm run build && firebase serve --only functions",
     "shell": "npm run build && firebase experimental:functions:shell",
     "start": "npm run shell",


nicolasgarnier:
Thanks @a-xin,

@inlined I confirm this issue, normally we should point to directory file rather than a `tslint.json` file in the `-p` option of `tslint`. See printout fromt `tslint --help`:

image

nicolasgarnier:
Since the file is auto-generated from `firebase init` we should probably change that.

typescript-getting-started/functions/package.json
```diff
@@ -11,7 +11,7 @@
     "typescript": "^2.6.1"
   },
   "scripts": {
-    "build": "./node_modules/.bin/tslint -p tslint.json && ./node_modules/.bin/tsc",
+    "build": "./node_modules/.bin/tslint --project tsconfig.json && ./node_modules/.bin/tsc",
```
nicolasgarnier: Actually it looks like this should be a directory so maybe:

```
"build": "./node_modules/.bin/tslint --project ./ && ./node_modules/.bin/tsc",
```

a-xin: @nicolasgarnier Yeah, the description of the cli parameter made me wonder about dir vs. file as well... I think they support both, most of the samples from their docs specify the file, so in the end I went for specifying the file instead of the dir. For example:
https://palantir.github.io/tslint/usage/type-checking/#cli
https://github.com/palantir/tslint/blame/master/src/tslint-cli.ts#L175

I'd go with specifying the file instead of the directory - what do you think?

a-xin:
@nicolasgarnier Here is the PR for firebase-tools
https://github.com/firebase/firebase-tools/pull/652

Btw, the documentation also references the wrong file:
https://firebase.google.com/docs/functions/typescript#using_an_existing_typescript_project

a-xin:
@nicolasgarnier https://github.com/firebase/firebase-tools/pull/652 got merged, what needs to be done here, ie. to get the file updated via `firebase init`?

And do you know what's the right place/channel to trigger an update to the docs as well?
https://firebase.google.com/docs/functions/typescript#using_an_existing_typescript_project

nicolasgarnier:
Sweet thanks @a-xin 

Let's merge this :)

I'll also start to change the docs internally (our docs are not on GitHub unfortunately). That will get published later.

nicolasgarnier:
Thanks again!

PR #322, opened 2018-02-09
Add scripts to package.json that now come with `firebase init`, incl linting

inlined:
This pull request is intentionally stacked so that robo-changes come before manual changes. I used git ls-files to find all package.json files and node to modify their JSON.

To test all files, I ran:

```bash
for i in $(git ls-files | grep 'functions/package.json'); do npm run lint --prefix $(dirname $i); done
```

The functions/ prefix was to detect cases where package.json was also used for firebase hosting.

Any NPM error meant that a `package.json` was missing the linter script (e.g. the code sample was added since I first robo-modified `package.json`s). Those were manually fixed. Any linter errors were also manually fixed.

### DIFF
--- a/assistant-say-number/functions/package.json
+++ b/assistant-say-number/functions/package.json
@@ -5,5 +5,17 @@
     "actions-on-google": "^1.0.7",
     "firebase-admin": "^4.0.5",
     "firebase-functions": "^0.5.1"
+  },
+  "devDependencies": {
+    "eslint": "^4.13.1",
+    "eslint-plugin-promise": "^3.6.0"
+  },
+  "scripts": {
+    "lint": "./node_modules/.bin/eslint .",
+    "serve": "firebase serve --only functions",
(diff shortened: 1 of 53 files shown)


nicolasgarnier:
Is this CL a good place to also cleanup the .eslintrc.json files:

 - remove the duplicate rule `"promise/always-return": 2,` appears twice
 - Extend the [Google lint rule](https://github.com/google/eslint-config-google) instead of  `"extends": "eslint:recommended",`  (we'd also need to import it in the devdependencies of all samples)

inlined:
That sounds like a good robochange. Let's do separate pull requests with robochanges and have each request document how the change was created.

inlined:
FWIW, we should also have a robochange that ensures firebase.json wires up the linter in predeploy rules.

nicolasgarnier:
cool LGTM then!

PR #323, opened 2018-02-09
Return a response so headers aren't set twice

nicoqh:
Otherwise the function will continue running after the 'Forbidden' response is sent.


nicoqh:
I signed it!

nicolasgarnier:
Thanks @nicoqh !

PR #327, opened 2018-02-17
remove return statement from console.log

tavelli:
causes compilation failure when using typescript

### DIFF
--- a/image-sharp/functions/index.js
+++ b/image-sharp/functions/index.js
@@ -89,5 +89,6 @@ exports.generateThumbnail = functions.storage.object().onChange((event) => {
 
   return streamAsPromise.then(() => {
     console.log('Thumbnail created successfully');
+    return null;
   });
 });
--- a/image-sharp/functions/index.js
+++ b/image-sharp/functions/index.js
@@ -88,6 +88,6 @@ exports.generateThumbnail = functions.storage.object().onChange((event) => {
     thumbnailUploadStream.on('finish', resolve).on('error', reject));
 
   return streamAsPromise.then(() => {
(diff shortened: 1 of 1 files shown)


tavelli:
i signed it

nicolasgarnier:
@tavelli Could you tell us a little more about the error that you are seeing with Typescript? (e.g. could you paste the error here?)

We added an ESLint rule to all projects that makes having a return statement inside promises mandatory to avoid some common errors that developer face when writing these return promises (e.g. forgetting to return a Promise or a chained/sub promise).

cc/ @inlined @tinaliang

nicolasgarnier:
FYI we added this as part of #306

tavelli:
@nicolasgarnier my bad i's actually tslint error not compile error, but it will cause deploy firebase deploy to fail if using default tslint config created during firebase init. 

the lint error is: 
ERROR: Expression has type `void`. Put it on its own line as a statement.

so it can also resolved by just moving return statement onto separate line after the console.log

```
return streamAsPromise.then(() => {
    console.log('Thumbnail created successfully');
    return;
})
```

inlined:
That looks good to me. I'd also accept inlined expressions as well (they're exempt from the always-return linter rule):

```
return streamAsPromise
    .then(() => console.log('Thumbnail created successfully'))
    .catch(err => console.error('Error creating thumbnail', err));
```

As some minor other fixes from my sample: always print your errors and print them to stderr.

nicolasgarnier:
I just added the `return null;` to  your PR. Let's merge!

Thanks @tavelli !

tavelli:
Thanks!

PR #333, opened 2018-03-02
Enable travis linting.

### LINKED ISSUE #325: Add testing, linting, and building
@nicolasgarnier would you be open to adding something like this:
https://github.com/firebase/snippets-node/pull/11

It's a few parts:

  * Use `lerna` to build and test each module
  * Lint all source files against a single `eslint` config to ensure correctness and consistency

If so I'd be happy to implement / assist implementation.

samtstern:
Fixes #325 

@nicolasgarnier


assistant-say-number/functions/package.json
```diff
@@ -11,7 +11,7 @@
     "eslint-plugin-promise": "^3.6.0"
   },
   "scripts": {
-    "lint": "./node_modules/.bin/eslint .",
+    "lint": "./node_modules/.bin/eslint --max-warnings=0 .",
```
samtstern: The `--max-warnings=0` is basically a strict mode to fail on warnings.

quickstarts/time-server/functions/index.js
```diff
@@ -56,7 +56,7 @@ exports.date = functions.https.onRequest((req, res) => {
 
   // [START usingMiddleware]
   // Enable CORS using the `cors` express middleware.
-  cors(req, res, () => {
+  return cors(req, res, () => {
```
samtstern: This is the one lint violation I found!

es2017-transpile/functions/index.js
```diff
@@ -0,0 +1,27 @@
+/**
```
samtstern: Any idea why this is here?  Should it be gitignored?

samtstern:
@nicolasgarnier note that I was not able to get `lerna bootstrap` to run on Travis in a reasonable about of time (I think because it spins up ~50 node processes) so I couldn't run the test.

If we can find a more efficient way to `npm install` in each module then we could run `lerna run test` to hit all of the tests.

We're able to do `lerna run lint` because that actually doesn't require installing anything besides the linter.

nicolasgarnier:
For `lerna run test` would it be possible to restrict it to only the modules that actually have tests? right now there is only one (quickstart > database) so we don't need to `npm install` on all the samples.

samtstern:
Ok these are all the other directories that contain `.js` and are not `functions`:

```
$ find . -type f -name "*.js" -not -path "*functions*" | xargs dirname | sort | uniq
./authenticated-json-api/public
./authorized-https-endpoint/public
./email-confirmation/public
./exif-images/public
./fcm-notifications/public
./fulltext-search-firestore/public
./instagram-auth/public
./isomorphic-react-app/src
./isomorphic-react-app/src/containers
./linkedin-auth/public
./nextjs-with-firebase-hosting/src/app
./nextjs-with-firebase-hosting/src/app/components
./nextjs-with-firebase-hosting/src/app/pages
./presence-firestore/public
./quickstarts/email-users/public
./spotify-auth/public
./text-moderation/public
./username-password-auth/public
```

I think it should be a separate issue to add an `.eslintrc` to each of those since I don't really feel qualified to make the style decisions there.

quickstarts/uppercase/functions/package.json
```diff
@@ -12,8 +12,9 @@
     "sinon": "^4.1.3"
   },
   "scripts": {
-    "test": "./node_modules/.bin/mocha --reporter spec",
-    "lint": "./node_modules/.bin/eslint .",
+    "build": "npm install",
```
samtstern: @nicolasgarnier got the tests running by adding this `build` step.

inlined: build should not do NPM install; that takes way too long and is needed so infrequently. It should be part of our travis rules.

samtstern: What I want is for "test" do depend on "install" so that when I do `lerna run test` it will run `npm install` in only the directories that have tests.

samtstern: How about I make a `travis-test` target that does `npm install && npm run test`

nicolasgarnier:
LGTM Thanks @samtstern !

We'll add the lin files for the frontend JS files separately as you suggested.

PR #340, opened 2018-03-09
Updated url-shortener to use api key not from firebase variable

prescottprue:
As of  [v3.17.5 of firebase-tools](https://github.com/firebase/firebase-tools/releases/tag/v3.17.5) `functions.config().firebase.apiKey` is no longer set therefore
a separate variable must be used or else RateLimit error may appear.


nicolasgarnier:
Just added some tweaks. Thanks @prescottprue for catching this!

PR #341, opened 2018-03-11
Fix #330

### LINKED ISSUE #330: "OAuth2Client is not a constructor" for google-sheet-sync
I am receiving the following error when attempting to deploy the google-sheet-sync function: 

> TypeError: OAuth2Client is not a constructor

The error points to line 45 in the index file:

`const functionsOauthClient = new OAuth2Client(CONFIG_CLIENT_ID, CONFIG_CLIENT_SECRET, FUNCTIONS_REDIRECT);`

I go through the 14 steps laid out exactly, configuring all the IDs and everything else. I believe the issue might also have to do with line 22 in the index file:

`const {OAuth2Client} = require('google-auth-library');`

Elsewhere (e.g. [on stack overflow here](https://stackoverflow.com/questions/48242920/google-calendar-api-and-node-js-googleauth-is-not-a-constructor-issue) and [on Google's GitHub here](https://github.com/google/google-auth-library-nodejs/issues/251)) people have talked about how the syntax for the OAuth2Client class and the GoogleAuth class was altered slightly, but this code reflects that change and this issue remains.

FluorescentHallucinogen:


nicolasgarnier:
Thanks @FluorescentHallucinogen !

PR #342, opened 2018-03-12
[google-sheet-sync] Fix return value error

FluorescentHallucinogen:
Fix `Error serializing return value: TypeError: Converting circular structure to JSON` error from Cloud Functions logs. This is the continuation of pull request #341.


nicolasgarnier:
Thanks again @FluorescentHallucinogen !

PR #371, opened 2018-04-10
Get user fields from data object

kroikie:
The user fields provided from an onCreate or onDelete event are inside of an internal data object.

To get the fields use: `user.data.<user field>` instead of `user.<user field>`


nicolasgarnier:
Hey @kroikie did you update to the latest Functions SDK? (v1.0.0) There was a breaking change. Looks like you are using the old syntax for the SDK < v1.0.0

PR #378, opened 2018-04-20
Exif function example, fix "not defined" errors.

sefk:
This example has been broken since this pull request on Feb 2018:
https://github.com/firebase/functions-samples/pull/306/files. Fixes "result not
defined" and "metadata not defined" errors but the overall control flow was
also pretty broken.

### DIFF
--- a/exif-images/functions/index.js
+++ b/exif-images/functions/index.js
@@ -47,18 +47,22 @@ exports.metadata = functions.storage.object().onFinalize((object) => {
   const bucket = gcs.bucket(object.bucket);
   return bucket.file(filePath).download({destination: tempLocalFile}).then(() => {
     // Get Metadata from image.
-    return spawn('identify', ['-verbose', tempLocalFile], {capture: ['stdout', 'stderr']});
-  }).then(() => {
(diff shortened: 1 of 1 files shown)


nicolasgarnier:
@sefk Thanks for the fix! I'll change this a bit since we need to follow a strict styling on this repo now and nested Promises are not allowed (which caused the original error to begin with).

@samtstern any idea why the travis build passed? I do see the ESLint errors in there

sefk:
Thanks for taking a look, @nicolasgarnier. Changed to unnest promises. Not sure if the state passing in line 57-58 is the best way to do that, feels kinda clunky.

Also glad that you're following up on why our automated tests didn't catch. Maybe other similar problems were introduced in [that same pull request](https://github.com/firebase/functions-samples/pull/306)?

nicolasgarnier:
Thanks for this @sefk ! Just tweaked it a bit (we need to return the `admin.database()...` promise to make sure the async jobs are chained) and also we enforce returning in every promise to avoid the mistake where developer forget to return Promises which result un functions terminating too early.

sefk:
Appreciate the cleanups! Thanks.

PR #380, opened 2018-04-23
 [isomorphic-react-app] fix broken user page #1

orouz:
on line 43, the assignment to `reportsId` is done using an undefined reference , and in any way the value is an object, so i believe  `Object.keys` is the way to go

on line 48, i changed the naming (`currentEmployee` -> `employee`) because in `User.jsx` on line `70` there's this: 

`const { employee } = this.props.currentEmployee`

so it breaks otherwise.

edit:

this should go with #381 too


nicolasgarnier:
Great fix @orouz thanks!

PR #396, opened 2018-05-11
Updated catch for req.cookie

jofftiquez:
The previous code will fail if no Authorization was set in the header.


nicolasgarnier:
Thanks @jofftiquez !

PR #404, opened 2018-05-16
Fix undefined variable

fvilers:
`data` is undefined (typo from previous implementation where `onUpdate` had an `event` sturcture). The typo is also present on the [documentation website](https://firebase.google.com/docs/firestore/solutions/presence).


nicolasgarnier:
Thanks @fvilers! good catch!

@egilmorez could you republish these docs after I;ve merged this?

fvilers:
@nicolasgarnier You're welcome!

PR #420, opened 2018-06-14
Use inclusive language

svenluijten:
His/hers can often be replaced with "their" to be more inclusive.


svenluijten:
I signed it!

svenluijten:
Travis fails but this PR doesn't seem to be the cause of that.

nicolasgarnier:
Thanks for the fix @svenluijten !

I'll investigate the issue with our travis build separately :) This is surely not the cause.

Issue #423, opened 2018-06-21
Instagram auth: Error in the token Function: undefined


alamboley:
Hello, I tried to clone the instagram auth exemple, but I've this issue when I add my instagram credentials: *Error in the token Function: undefined*.

My app is in sandbox, and my user added.

```js
exports.token = functions.https.onRequest((req, res) => {
  const oauth2 = instagramOAuth2Client();

  try {
    return cookieParser()(req, res, () => {
      console.log('Received verification state:', req.cookies.state);
      console.log('Received state:', req.query.state);
      if (!req.cookies.state) {
        throw new Error('State cookie not set or expired. Maybe you took too long to authorize. Please try again.');
      } else if (req.cookies.state !== req.query.state) {
        throw new Error('State validation failed');
      }
      console.log('Received auth code:', req.query.code);
      oauth2.authorizationCode.getToken({
        code: req.query.code,
        redirect_uri: OAUTH_REDIRECT_URI,
      });
    }).then((results) => {
      console.log('Auth code exchange result received:', results);
```
Thanks to Firebase cloud functions diary, I see the `Received auth code` log, but nothing after that. Sounds like the `oauth2.authorizationCode.getToken` fails.

What did I miss?

Thanks for the help!

alamboley:
Up @nicolasgarnier ?

samtstern:
I believe this was fixed by #424

PR #424, opened 2018-06-25
return a correct error not undefined like in #423

alamboley:


alamboley:
I signed it!

nicolasgarnier:
Thanks @alamboley !

PR #427, opened 2018-06-30
authorized-https-endpoint: getToken deprecated

ppannuto:
The `getToken` method has been deprecated and replaced with `getIdToken`.

I've verified that the example runs correctly locally after this change and now produces no warnings.


nicolasgarnier:
Sweet, thanks @ppannuto ! :)

PR #437, opened 2018-07-24
Samples updated to ES2017 Syntax and Node 8 runtime

nicolasgarnier:
This branch contains all samples updated to use ES2017 syntax and Node 8 as the runtime.

We're not merging to master yet as Node 6 is still the default runtime.


nicolasgarnier:
Note about the bot warning above: the commits from different authors are from PRs so their Google CLA status has been checked in that PR.

PR #441, opened 2018-08-02
clear _session cookie on signOut

rpidanny:
The API could still be accessed after sign out as the _session cookie still contains the user token.


nicolasgarnier:
Thanks @rpidanny !

PR #442, opened 2018-08-03
rm deprecated crypto module, fix grammar

dandv:
https://www.npmjs.com/package/crypto

+ move cron-jobs to the top since it's free


nicolasgarnier:
Thanks @dandv !

PR #444, opened 2018-08-09
Return firebase auth token after creating account in the linkedin-auth sample

cxam:
The firebase auth token from the `createFirebaseAccount` function in the linkedin-auth sample is not returned and causes the following client error: `Error in the token Function: undefined`. This PR fixes the function to return the token.


nicolasgarnier:
LGTM thanks @cxam

PR #452, opened 2018-08-20
Fix outdated code in text-moderation sample

krawaller:
Currently the code in the `text-moderation` sample throws an error as `snap.val` doesn't exist, since `onWrite` now receives a `change` object instead of a `snapshot`.


krawaller:
CLA signed.

nicolasgarnier:
Thanks @krawaller !

PR #462, opened 2018-09-09
Update getToken from functinosOauthClient

luskin:
The sample code shows to set the results of `getToken` directly to the variable tokens however tokens is now a property of the response so you need to store the tokens object from within the response as google states here:

https://github.com/google/google-api-nodejs-client:
```
const {tokens} = await oauth2Client.getToken(code)
oauth2Client.setCredentials(tokens);
```


luskin:
I have signed it!

nicolasgarnier:
Thanks @luskin !

PR #473, opened 2018-09-15
Fix instagram-auth sample

romanalexander:
Misplaced promise syntax on cookieParser rather than getToken


nicolasgarnier:
good catch. Thanks @romanalexander !

PR #538, opened 2019-02-03
Update index.js

rishitoshsingh:
I was getting a TypeError: "snapshot.childCount is not a function"
What's wrong: numChildren() is a DataSnapshot interface but snapshot is not a DataSnapshot.

I've fixed this error.

### DIFF
--- a/limit-children/functions/index.js
+++ b/limit-children/functions/index.js
@@ -26,11 +26,11 @@ exports.truncate = functions.database.ref('/chat').onWrite((change) => {
   const parentRef = change.after.ref;
   const snapshot = change.after
 
-  if (snapshot.numChildren() >= MAX_USERS) {
+  if (snapshot.numChildren() >= MAX_LOG_COUNT) {
     let childCount = 0;
     const updates = {};
     snapshot.forEach((child) => {
-      if (++childCount <= snapshot.numChildren() - MAX_USERS) {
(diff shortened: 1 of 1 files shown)


rishitoshsingh:
I signed it!!

limit-children/functions/index.js
```diff
@@ -22,14 +22,15 @@ const MAX_LOG_COUNT = 5;
...
 // In this example we'll keep the max number of chat message history to MAX_LOG_COUNT.
-exports.truncate = functions.database.ref('/chat/{messageid}').onWrite(async (change) => {
-  const parentRef = change.after.ref.parent;
-  const snapshot = await parentRef.once('value');
-  if (snapshot.numChildren() >= MAX_LOG_COUNT) {
+exports.truncate = functions.database.ref('/chat').onWrite((change) => {
+  const parentRef = change.after.ref;
+  const snapshot = change.after
+
+  if (snapshot.numChildren() >= MAX_USERS) {
```
jhuleatt: What prompted the switch from `MAX_LOG_COUNT` to `MAX_USERS`? Looks like the comment in lines 23 and 24 show that this function was intentionally using `MAX_LOG_COUNT`

jhuleatt:
A very belated thank you for this fix, @rishitoshsingh!

PR #548, opened 2019-03-04
Updated  for recent cloud-vision changes

oldirty91:
The original index.js code that interfaced with cloud-vision would not function. Checked the cloud-vision docs and they did not align with what was originally here. I've updated the code to what I was able to get  working.


oldirty91:
I signed it!

jhuleatt:
[cleaning up old issues]

Thanks for this PR @oldirty91 and sorry it was never accepted. The [current code for moderate-images](https://github.com/firebase/functions-samples/blob/master/moderate-images/functions/index.js) does use the newer vision libraries

PR #576, opened 2019-06-10
Fixed auth middleware

rahulvyas1:
The auth middleware had an issue as the decodedIdToken was not being passed to the routes.

So, I used the claims object that was returned after decoding the Id token and then passed that as the user for req.user

Cheers


rahulvyas1:
I signed it!

jhuleatt:
Thanks for taking the time to submit this, but I'm going to close this one. #596 fixes a number of bugs in the `authenticated-json-api` sample.

PR #605, opened 2019-08-21
wait for resolving of promise

byildiz:
verifyIdToken returns a promise (https://firebase.google.com/docs/reference/admin/node/admin.auth.Auth.html#verifyidtoken). To use decodedIdToken we should wait for resolving.


byildiz:
@googlebot I signed it!

jhuleatt:
Thanks @byildiz!

PR #609, opened 2019-08-28
Exif function example, fix "metadata.downloadURLs" deprecated

mxlius:
Fix error message: "Cannot read property '0' of undefined"
Root cause: Storage API ```FullMetadata.downloadURLs``` is deprecated, ref:https://firebase.google.com/docs/reference/js/firebase.storage.FullMetadata.html#downloadurls

Replace with ```Reference.getDownloadURL```, https://firebase.google.com/docs/reference/js/firebase.storage.Reference.html#getdownloadurl


mxlius:
@googlebot I signed it!

mxlius:
@googlebot I signed it!

ramblescout:
Adding a note that I couldn't get this to work (node 12).  I kept getting an error that downloads[0] wasn't an object. I switched back to the "deprecated" version and now it works.  Might have trouble later, if so I'll come back and update.

Issue #594, opened 2019-08-06
Wish you could create one sample on this: The function using nodejs-dialogflow: https://github.com/googleapis/nodejs-dialogflow


eGlobeBizCom:
Hi All,

I believe that this sample will be useful to lots of people who want to use this module:
https://github.com/googleapis/nodejs-dialogflow

Thanks

PR #611, opened 2019-09-06
Dialogflow API use sample added

krharsh17:
In reference with issue #594. The sample uses the functions platform to send requests to and receive responses from the dialogflow api. (www.dialogflow.com)


krharsh17:
@googlebot I signed it!

PR #648, opened 2019-12-19
add new cloud function

makarushka:
Function that triggers user firestore
and after the user has changed avatar id, the function checks: that such an id is in the user's photo collection or a non-empty id is set, otherwise it changes to the previous identifier


jhuleatt:
@makarushka thanks for the PR, but since we already have a Firestore-triggered example (uppercaser), we're not going to accept this one

PR #660, opened 2020-01-03
Change the dead link of Sharp

zawataki:


samtstern:
@zawataki thanks for this!

PR #665, opened 2020-01-19
Update Stripe Package

theswerd:
The current package.json uses a stripe package that is more than 3 years old. I want to start updating the cloud functions, and the first step to doing that is adding the newest functionality.


jhuleatt:
Thanks @theswerd !

PR #679, opened 2020-02-28
fix package.json files to un-bork DPEBot

jhuleatt:
The package.json files containing `"next": "beta"` were causing dpebot to crash every time it tried to run. I fixed those files and ran dpebot's script manually to get everything back up to date.


arifmyself:
> The package.json files containing `"next": "beta"` were causing dpebot to crash every time it tried to run. I fixed those files and ran dpebot's script manually to get everything back up to date.

Issue #684, opened 2020-03-06
Deployed but cannot generate thumbnail on uploaded images


defeinium:
I have tried to follow all of the steps and the function is deployed. However, the thumbnail is not generated when I upload an image and the real time database is showing null. Is it that any steps I missed out >

jhuleatt:
Hi @defeinium, I'm able to reproduce this. It is caused by the npm dependency [`mkdirp-promise`](https://github.com/ahmadnassri/mkdirp-promise/issues/89). I'm working on swapping that out `mkdirp` now.

qlereboursBS:
I had the same problem, the timeout is reached because of mdkirp-promise.
I will switch too

dungkaka:
@jhuleatt , i wrote my function to generate thumbnail by following each step. Now I have trouble like this issue post, I try to fix that by replace mkdirp but nothing change. Thumbnail still can not be generated. 
    "firebase": "^7.10.0",
    "firebase-admin": "^8.6.0",
    "firebase-functions": "^3.3.0",
    "child-process-promise": "^2.2.1",
    "mkdirp": "^1.0.3",
Actually, thumbnail appearred on my bucket, but just loading icon on image.
Can you check it ?

jhuleatt:
The loading icon issue is being discussed in #652

PR #687, opened 2020-03-11
Fix generate-thumbnail: swap out mkdirp-promise

jhuleatt:
[mkdirp](https://www.npmjs.com/package/mkdirp) is now promise-based, and [mkdirp-promise](https://www.npmjs.com/package/mkdirp-promise) no longer works. mkdirp-promise caused this function to hang forever when it tried to create a local directory.


jhuleatt:
fixes #684

PR #705, opened 2020-04-22
fix: lock sharp version

Salakar:
Locks the Sharp library version to a Node.js v8 compatible version. 

See https://github.com/firebase/extensions/pull/278 for details


jhuleatt:
Thanks @Salakar !

PR #707, opened 2020-04-24
Fix npm build warning

### LINKED ISSUE #690: "npm build" instead of "npm run build"
https://github.com/firebase/functions-samples/blob/8c5e715870fa975a027e7bf65bac1af915f05655/nextjs-with-firebase-hosting/package.json#L14

Pretty sure that should be `npm run build` not `npm build`. 
Currently it throws a warning from npm, whereas npm run build actually does build next.

dlackty:
This closes #690. I also do a quick search for `npm build` across the whole project, and this is the only usage.


jhuleatt:
Thank you @dlackty !

PR #710, opened 2020-04-30
Add sample to publish models

ulukaya:


publish-model/functions/index.js
```diff
@@ -0,0 +1,86 @@
...
+  
+  const modelOptions = {
+    displayName: modelName,
+    tfliteModel: { gcsTfliteUri: gcsUri }
+  };
+  
+  // List models with same name
+  const modelsResult = await ml.listModels({filter: `display_name = ${modelName}`});
+  
+  var model;
```
jhuleatt, line 62: since you're using `const` everywhere else, you can use `let` here. but not a big deal either way

publish-model/functions/index.js
```diff
@@ -0,0 +1,86 @@
...
+  } else {
+	  model = await ml.createModel(modelOptions);
+  }
+
+  if (model.validationError) {
+    // Can't publish models with validation errors
+    return console.log(`Validation error: ${model.validationError}. Will not publish the model.`);
+  }
+
+  const publishedModel = await ml.publishModel(model.modelId);
```
jhuleatt, line 83: you might get linter errors here for defining a variable and not using it. you can just make this line `await ml.publishModel(model.modelId);` or if there's something in the `publishedModel` object that's worth including in the final log, you can add it there. also not a big deal either way

PR #718, opened 2020-05-13
Call docChanges method according to TS definition

bodazhao:
TypeScript warned me this error when chaining `forEach` after `docChanges`

```
Property 'forEach' does not exist on type '(options?: SnapshotListenOptions |
undefined) => DocumentChange<DocumentData>[]'.ts(2339)
```

So I look at [the doc](https://firebase.google.com/docs/reference/js/firebase.firestore.QuerySnapshot#docchanges) and see it's an method and by calling it fixed the ts error


bodazhao:
@googlebot I signed it!

jhuleatt:
Thank you for finding this bug @bodazhao !

PR #722, opened 2020-05-19
updated mkdirp-promise import

mauzepeda:
`mkdirp-promise` is deprecated so I changed it to `mkdirp` as suggested in the official documentation. It should be backwards compatible.


jhuleatt:
Thanks @mauzepeda!

PR #728, opened 2020-05-31
Update Stripe Sample

thorsten-stripe:
* Update to new Stripe Payments APIs and Stripe Elements
* Change to Vanilla HTML and JavaScript
* Update Stripe libraries
* ~Change to node 10 engine~

You can test it here: https://cloud-functions-stripe-sample.web.app/
- no 3D Secure required: 4242424242424242
- 3D Secure required: 4000000000003220


stripe/public/javascript/app.js
```diff
@@ -0,0 +1,319 @@
...
+    .collection('stripe_customers')
+    .doc(currentUser.uid)
+    .collection('payment_methods')
+    .onSnapshot((snapshot) => {
+      if (snapshot.empty) {
+        document.querySelector('#add-new-card').open = true;
+      }
+      document
+        .querySelectorAll('select[name=payment-method] option')
+        .forEach((el) => el.remove());
```
thorsten-stripe, line 103: @jhuleatt is there a better way of doing this? I did try the `snapshot.docChanges()`[0] approach and tried to filter for `change.type === 'added'` but that didn't behave as expected and also appended too many items.

I there a way to only get the item that was added?

[0] https://firebase.google.com/docs/firestore/query-data/listen#view_changes_between_snapshots

thorsten-stripe:
@googlebot I signed it!

thorsten-stripe:
@googlebot I fixed it.

thorsten-stripe:
@jhuleatt any idea what's going wrong here? It looks like the bot doesn't like the merging in of the changes from master. Any idea why that is?

jhuleatt:
It looks like the CLA bot is confused. I'm going to manually set the CLA label

stripe/public/javascript/app.js
jhuleatt: The way I usually do it is something like:

```js
snapshot.forEach(doc => {
  const paymentMethod = doc.data();
  if (!paymentMethod.card) {
    return;
  }

  const optionId = `card-${doc.id}`;
  let optionElement = document.getElementById(optionId);

  // add a new option if one doesn't exist yet
  if (!optionElement) {
    optionElement = document.createElement("option");
    optionElement.id = optionId;
    document.querySelector("select[name=payment-method]").appendChild(option);
  }

  optionElement.text = `${paymentMethod.card.brand} •••• ${paymentMethod.card.last4} | Expires ${paymentMethod.card.exp_month}/${paymentMethod.card.exp_year}`;
});
```

stripe/functions/index.js
```diff
@@ -18,77 +18,142 @@
...
-        // still logging an exception with StackDriver
-        console.log(error);
-        await snap.ref.set({error: userFacingMessage(error)}, { merge: true });
-        return reportError(error, {user: context.params.userId});
-      }
-    });
-// [END chargecustomer]]
+const { Logging } = require('@google-cloud/logging');
+const logging = new Logging({
+  projectId: '<YOUR FIREBASE PROJECT ID>',
```
jhuleatt, line 23: Instead of asking a developer to set this manually, this can be accessed as an environment variable that's automatically provided to Cloud Functions via `process.env.GCLOUD_PROJECT`

stripe/functions/package.json
```diff
@@ -20,7 +20,7 @@
     "logs": "firebase functions:log"
   },
   "engines": {
-    "node": "8"
+    "node": "10"
```
jhuleatt, line 23: Let's leave this on 8 for now so all samples in this repo are aligned. There are some changes coming with Node 10 that we need to address across this whole repo. [Background](https://firebase.google.com/support/faq/#expandable-15)

jhuleatt:
Added some small comments, but looks good to me overall!

stripe/public/javascript/app.js
```diff
@@ -0,0 +1,319 @@
...
+      .firestore()
+      .collection('stripe_customers')
+      .doc(currentUser.uid)
+      .onSnapshot((snapshot) => {
+        if (snapshot.data()) {
+          customerData = snapshot.data();
+          startDataListeners();
+          document.getElementById('loader').style.display = 'none';
+          document.getElementById('content').style.display = 'block';
+        }
```
jhuleatt, line 61: While we're here, let's fix a small bug. We have an `if (snapshot.data())`, but can you please add an `else` that console.logs or similar to indicate that no customer record for this user ID was found? I ran into this because I signed in with a user that had already been created on my project, so the auth-triggered function never ran.

stripe/functions/index.js
```diff
@@ -18,77 +18,142 @@
...
 
-// Add a payment source (card) for a user by writing a stripe payment source token to Cloud Firestore
-exports.addPaymentSource = functions.firestore.document('/stripe_customers/{userId}/tokens/{pushId}').onCreate(async (snap, context) => {
-  const source = snap.data();
-  const token = source.token;
-  if (source === null){
-    return null;
-  }
+/**
+ * When adding the payment method ID on the clinet,
```
jhuleatt, line 47:
```suggestion
 * When adding the payment method ID on the client,
```

thorsten-stripe:
Thanks a mill for the review, I'll look into the needed changes :)

Regarding the `No such setupintent` error, it looks like you're using the default publishable key from the Stripe docs, rather than your Stripe account's publishable key. Updating the `STRIPE_PUBLISHABLE_KEY` in the `app.js` file with the one in your Stripe Dashboard should do the trick 👍 

To help others avoid this mistake, do you remember where you copied the publishable key from?

stripe/public/javascript/app.js
thorsten-stripe: Brilliant, I've changed to that approach, thanks :)

stripe/functions/package.json
thorsten-stripe: These changes mainly seem related to billing, right? Since we're making requests to Stripe we need to enable billing anyway, no?

jhuleatt:
> Thanks a mill for the review, I'll look into the needed changes :)
> 
> Regarding the `No such setupintent` error, it looks like you're using the default publishable key from the Stripe docs, rather than your Stripe account's publishable key. Updating the `STRIPE_PUBLISHABLE_KEY` in the `app.js` file with the one in your Stripe Dashboard should do the trick 👍
> 
> To help others avoid this mistake, do you remember where you copied the publishable key from?

Thanks Thor, that was it! I copy + pasted from a draft update of [this page](https://firebase.google.com/docs/use-cases/payments) that will go along with this PR, and forgot to replace the default key with my own. I'll make a note on the draft to use something like `<YOUR STRIPE PUBLISHABLE KEY>` instead of the default publishable key.

PR #730, opened 2020-06-02
Update sharp, remove deprecated max fn call, retrieve image urls and …

alxvallejo:
…save to firebase rtdb.

Updating the example here on how to better use the sharp node lib to parse and save to the firebase db.

Sorry for the formatting - i'm actually have some minor issue with VSCode on that. If anyone can assist there that'd be great.

### DIFF
--- a/image-sharp/functions/index.js
+++ b/image-sharp/functions/index.js
@@ -16,9 +16,12 @@
 'use strict';
 
 const functions = require('firebase-functions');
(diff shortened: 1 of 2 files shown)


NickFoden:
Hi @alxvallejo in vscode, you can  "Command + Shift + P" to select a save without formatting, this might help you reduce the amount of whitespace, etc  changes?

PR #734, opened 2020-06-07
add Test for onRequest function

sensuikan1973:
## Overview
add _Simple_ Cloud Functions Unit Test with TS.
Regards,

## Reference
* implement
  * https://firebase.google.com/docs/functions/unit-testing#testing_http_functions
* dependency: use `mocha` and `chai`
  * https://github.com/firebase/functions-samples/blob/master/quickstarts/uppercase/functions/package.json
  * https://github.com/firebase/quickstart-nodejs/pull/124

### DIFF
--- a/typescript-getting-started/functions/test/test.ts
+++ b/typescript-getting-started/functions/test/test.ts
@@ -1,0 +1,21 @@
+import { suite, test } from '@testdeck/mocha'
+import * as chai from 'chai'
+import * as functions from 'firebase-functions'
+
+import { helloWorld } from '../src'
+
+const assert = chai.assert;
+
+@suite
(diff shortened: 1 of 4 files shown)


typescript-getting-started/functions/test/test.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import { helloWorld } from '../src'
+
+const assert = chai.assert;
+
+@suite
+class HelloWorld {
+  @test
+  async helloWorld() {
+    const req = {} as functions.https.Request;
```
sensuikan1973, line 13: used `as`.
It's because I don't add mock library for only this and this test is very small and simple.

PR #743, opened 2020-06-26
Remove useless comments

Sxribe:
This file was cluttered with absolutely useless comments that did nothing except make the file hard to read. This file is one of the few examples that shows how to use cors with firebase functions.

For example:
```js
// [START readBodyParam]
format = req.body.format;
 // [END readBodyParam]
```
This code snippet provides **zero** additional information, however it clutters the code.

Starting the file with `[START all]` and ending with `[END all]` is once again an example of a useless comment. It's obvious that the code starts there, and ends there.

Comments should be used only for documentation, clarification, and nuances/things that need to be pointed out to future contributors.

### DIFF
--- a/quickstarts/time-server/functions/index.js
+++ b/quickstarts/time-server/functions/index.js
@@ -15,19 +15,15 @@
  */
 'use strict';
 
-// [START functionsimport]
 const functions = require('firebase-functions');
-// [END functionsimport]
-// [START additionalimports]
+
 // Moments library to format dates.
 const moment = require('moment');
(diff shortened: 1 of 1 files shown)


Sxribe:
@googlebot I signed it!

jhuleatt:
@Sxribe these comments are used by a script that inserts code from these samples into the Firebase docs, so we can't remove them. Apologies for the clutter.

Sxribe:
Makes sense, thanks

PR #767, opened 2020-09-24
docs: remove all references of the Flame plan

rosariopfernandes:
Removing all references about the flame plan from our docs, since it is no longer offered by Firebase.


jhuleatt:
Thank you @rosariopfernandes! Longer term, we should have shared language across the samples [once Node 8 is deprecated](https://firebase.google.com/support/faq?hl=sr#expandable-10), but this is a very helpful stopgap until then.

PR #787, opened 2020-11-12
Migrate from mkdirp-promise to mkdirp

Zeb88:
This package is broken and no longer maintained. 'mkdirp' itself supports promises now. Not switching to this will cause timeouts in blur function.


Zeb88:
@googlebot I signed it!

jhuleatt:
Thanks @Zeb88

PR #788, opened 2020-11-13
Add vision-annotate-images sample function

nauthadr:


vision-annotate-images/README.md
```diff
@@ -0,0 +1,16 @@
+# Vision Annotate Images
+
+This function acts as a pass through to the [Cloud Vision JSON API](https://cloud.google.com/vision/docs/request). 
+
+To call this from an app, you must first sign-in using [Firebase Auth](https://firebase.google.com/docs/auth).
+Only users who pass a valid Firebase ID token as a Bearer token in the `Authorization` header of the HTTP request are authorized to use the function.
+
```
nauthadr, line 7: I'm thinking once the firebase docs are ready we can just link to those as an example of how to use this in an app. What do you think?

samtstern: Yep sg

vision-annotate-images/functions/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "start": "npm run shell",
+    "deploy": "firebase deploy --only functions",
+    "logs": "firebase functions:log"
+  },
+  "engines": {
+    "node": "10"
+  },
+  "main": "lib/index.js",
+  "dependencies": {
+    "firebase-admin": "^8.10.0",
```
samtstern, line 17: `firebase-admin` 9+ is the version to use when using Node 10+ (which you are, according to the `engines` field)

vision-annotate-images/functions/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "firebase-functions": "^3.6.1",
```
samtstern, line 18: This looks a little old as well.

vision-annotate-images/functions/src/index.ts
```diff
@@ -0,0 +1,29 @@
+import * as functions from "firebase-functions";
+import * as admin from "firebase-admin";
+import vision from "@google-cloud/vision";
+
+admin.initializeApp();
```
samtstern, line 5: nit: doesn't look like you actually use `firebase-admin` at all in this function?

vision-annotate-images/functions/package.json
nauthadr: Done

vision-annotate-images/functions/package.json
nauthadr: Updated to 3.11.0

samtstern (approved):

vision-annotate-images/functions/src/index.ts
nauthadr: You're right. For some reason I thought it was required to do initializeApp. I've removed it.

PR #806, opened 2020-12-18
add deleting payments on cleanupUser function

### LINKED ISSUE #792: I think there should be code deleting 'payments' subcollection in 'cleanupUser' function
Thank you for these really nice samples. 
I just found a tiny omission in the code 

I think there should be code deleting 'payments' subcollection in 'cleanupUser' function 
just like this 
```
 const paymentsSnapshot = await dbRef.doc(user.uid).collection('payments').get()
    paymentsSnapshot.forEach((snap)=> snap.ref.delete())
```

Please let me know if there was an intention on it. 

https://github.com/firebase/functions-samples/blob/561d73b0357f4b03f4952242ee9513d57708f068/stripe/functions/index.js#L144-L148

HYJ29:
It closes #792. There is a tiny omission on the 'cleanupUser' function. 

So I just added code deleting all payment records as it said on the comment.  
(I thought about just fixing the comment but I think this way is more reasonable)

### DIFF
--- a/stripe/functions/index.js
+++ b/stripe/functions/index.js
@@ -144,11 +144,17 @@ exports.cleanupUser = functions.auth.user().onDelete(async (user) => {
   const customer = (await dbRef.doc(user.uid).get()).data();
   await stripe.customers.del(customer.customer_id);
   // Delete the customers payments & payment methods in firestore.
-  const snapshot = await dbRef
+  const paymetsMethodsSnapshot = await dbRef
     .doc(user.uid)
     .collection('payment_methods')
     .get();
-  snapshot.forEach((snap) => snap.ref.delete());
+  paymetsMethodsSnapshot.forEach((snap) => snap.ref.delete());
+  const paymentsSnapshot = await dbRef
+    .doc(user.uid)
+    .collection('payments')
+    .get();
+  paymentsSnapshot.forEach((snap) => snap.ref.delete());
+
   await dbRef.doc(user.uid).delete();
   return;
 });
--- a/stripe/functions/index.js
+++ b/stripe/functions/index.js
@@ -144,16 +144,19 @@ exports.cleanupUser = functions.auth.user().onDelete(async (user) => {
   const customer = (await dbRef.doc(user.uid).get()).data();
   await stripe.customers.del(customer.customer_id);
   // Delete the customers payments & payment methods in firestore.
+  const batch = admin.firestore().batch();
   const paymetsMethodsSnapshot = await dbRef
     .doc(user.uid)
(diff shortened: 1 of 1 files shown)


samtstern:
@HYJ29 thanks for the contribution! Assigning to @jhuleatt for review,

stripe/functions/index.js
```diff
@@ -144,11 +144,17 @@ exports.cleanupUser = functions.auth.user().onDelete(async (user) => {
   const customer = (await dbRef.doc(user.uid).get()).data();
   await stripe.customers.del(customer.customer_id);
   // Delete the customers payments & payment methods in firestore.
-  const snapshot = await dbRef
+  const paymetsMethodsSnapshot = await dbRef
     .doc(user.uid)
     .collection('payment_methods')
     .get();
-  snapshot.forEach((snap) => snap.ref.delete());
+  paymetsMethodsSnapshot.forEach((snap) => snap.ref.delete());
```
jhuleatt, line 151: I'm realizing that we mistakenly weren't `await`-ing the delete here. That means that this function won't reliably delete all information. I believe we'll need to do something like:

```js
const deletes = [];

paymetsMethodsSnapshot.forEach((snap) => deletes.push(snap.ref.delete()));

// more code ...

paymentsSnapshot.forEach((snap) => deletes.push(snap.ref.delete()));

// the important part
await Promise.all(deletes);
```

HYJ29: Yes, I totally agree with you. I missed that too.. thank you!
I think using an array is more simple than using `batch` as instructed in docs https://firebase.google.com/docs/firestore/manage-data/delete-data
how do you think about using batch though like below.
Is there any reason you used just array? 
```
const batch = adminfirestore().batch();
  const paymetsMethodsSnapshot = await dbRef
    .doc(user.uid)
    .collection('payment_methods')
    .get();
  paymetsMethodsSnapshot.forEach((snap) => batch.delete(snap.ref));
  const paymentsSnapshot = await dbRef
    .doc(user.uid)
    .collection('payments')
    .get();
  paymentsSnapshot.forEach((snap) => batch.delete(snap.ref));

  await batch.commit();
```

jhuleatt: Happy New Year @HYJ29! Great point, I think a `batch` delete definitely makes sense so that we can be confident all data was wiped out at once.

HYJ29: Ok. I added it as said above! Happy new year! @jhuleatt

jhuleatt (approved):

jhuleatt:
Thank you @HYJ29!

PR #811, opened 2020-12-29
show full error error like #423

OfJunker:
update spotify-auth to show full error instead of just undefined like in PR #423


OfJunker:
@googlebot I signed it!

PR #812, opened 2020-12-29
give full error like #423

OfJunker:
update spotify auth to give full error instead of just undefined, same as in PR #423


samtstern:
@OfJunker nice catch and thanks for the PR!

PR #826, opened 2021-02-10
Fix OAuth2.0 based authentication examples problems using cookie

nokazn:
Some examples using OAuth 2.0 based authentication with cookie doesn't work correctly in modern browser like Google Chrome and Firefox.
In these examples, `state` value are not set correctly in cookie because the requests for Cloud Functions are sent in cross domain and are not Top Level Navigation.
If `SameSite` attribute are not set, browsers treats it as `Lax` value by default.  So I think `SameSite` attributes should be set to `None`.

### DIFF
--- a/instagram-auth/functions/index.js
+++ b/instagram-auth/functions/index.js
@@ -64,6 +64,7 @@ exports.redirect = functions.https.onRequest((req, res) => {
       maxAge: 3600000,
       secure: true,
       httpOnly: true,
+      sameSite: 'none',
     });
     const redirectUri = oauth2.authorizationCode.authorizeURL({
       redirect_uri: OAUTH_REDIRECT_URI,
--- a/linkedin-auth/functions/index.js
+++ b/linkedin-auth/functions/index.js
@@ -56,6 +56,7 @@ exports.redirect = functions.https.onRequest((req, res) => {
       maxAge: 3600000,
       secure: true,
       httpOnly: true,
+      sameSite: 'none',
     });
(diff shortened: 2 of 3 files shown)


samtstern:
@nokazn thanks for this contribution! I don't know enough about browser cookies to review it but I hope @jhuleatt does.

kevinguebert:
@nokazn I ran into the same issues with authentication (for spotify) and tried out this solution. When setting setting `sameSite: 'none'` I received this error that caused the function not to run: 
```
TypeError: option sameSite is invalid
    at Object.serialize (/workspace/node_modules/cookie/index.js:174:15)
    at ServerResponse.res.cookie (/workspace/node_modules/express/lib/response.js:853:36) 
```

Changing it up to `SameSite: 'none'` with the first `S` uppercased solved the error. Have you seen that issue at all?

nokazn:
@kevinguebert
> Changing it up to `SameSite: 'none'` with the first `S` uppercased solved the error. Have you seen that issue at all?

I tried some times, but couldn't reproduce the above error in my environment.
I think it's a dependency problem. The version of `cookie` package which `express` in `firebase-functions`, or `cooike-parser` uses may be old. `cookie@0.3.x` cannot address `sameSite: 'none'` attribute.
You can also check [here](https://github.com/firebase/firebase-functions/issues/653).

mjroyappa:
why is `sameSite: 'None'` still not part of the latest sample? I was stuck on this for quite a while thinking my setup was wrong or something

PR #829, opened 2021-02-17
Changed getLastSignInTimestamp to getLastRefreshTimestamp in detectio…

TarikHuber:
Changed `getLastSignInTimestamp` to `getLastRefreshTimestamp` in detection of inactive users because `getLastSignInTimestamp` is only changed when user signs in and `getLastRefreshTimestamp` shows the real last active timestamp. If the user doesn't manuelly logout and login again the `getLastSignInTimestamp` can very easiely get older then one month and we get a lot active users deleted.

### DIFF
--- a/delete-unused-accounts-cron/functions/index.js
+++ b/delete-unused-accounts-cron/functions/index.js
@@ -13,12 +13,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-'use strict';
+"use strict";
 
-const functions = require('firebase-functions');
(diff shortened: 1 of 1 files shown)


jhuleatt:
Thanks for improving this sample @TarikHuber!

jhuleatt:
I ended up merging #945 which is a smaller change that accomplishes a similar goal and falls back to `getLastSignInTimestamp` when `getLastRefreshTimestamp` is not available.

Sorry for the delay here, but I'm going to close this out now that #945 is merged.

PR #836, opened 2021-03-03
Switch all samples from console.log to functions.logger.log

jhuleatt:
Use the [functions logger SDK recommended in the docs](https://firebase.google.com/docs/functions/writing-and-viewing-logs?hl=en#logger-sdk) instead of `console.log`.

I did a find+replace and did my best to catch edge cases:

- any in-browser JS needs to keep `console.log`
- links to the firebase console (`https://console.firebase.google.com`) shouldn't be changed to `https://functions.logger.firebase.google.com` 😆

### DIFF
--- a/authenticated-json-api/functions/index.js
+++ b/authenticated-json-api/functions/index.js
@@ -62,7 +62,7 @@ app.use(authenticate);
 app.post('/api/messages', async (req, res) => {
   const message = req.body.message;
 
-  console.log(`ANALYZING MESSAGE: "${message}"`);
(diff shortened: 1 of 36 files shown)


jhuleatt:
looks like the handlebars template failed the CI. Need to import functions.logger in firebaseUser.js: https://github.com/firebase/functions-samples/blob/05ce895949ec686bf034ac28416e8a2783537261/template-handlebars/functions/firebaseUser.js#L18-L27

samtstern (approved):

jhuleatt:
Checks are passing now!

PR #847, opened 2021-03-19
Fix ffmpeg path access

SudoSharma:
I tried running this function to convert audio files to the .flac format and received an error that the ffmpeg path was not found. It appears as if the `ffmpeg_static.path` should actually be `ffmpeg_static` as mentioned here in the [documentation](https://www.npmjs.com/package/ffmpeg-static). 

I've verified that this change works and actually does help convert audio files to .flac. 

Please let me know if I need to do anything else for this pull request and thank you for the awesome samples! They really help accelerate projects.


jhuleatt:
Thank you @SudoSharma!

jhuleatt (approved):

PR #849, opened 2021-03-31
Improve cookie issues in Instragram & LinkedIn samples

tomcurran:
Makes cookies first party by running the Firebase functions 'redirect' & 'token' through the same domain as the Firebase Hosting site. Similar idea to #826 but without changing the cookies to SameSite:None.


tomcurran:
@googlebot I signed it!

samtstern:
@tomcurran I like it! Thanks for the contribution.

PR #864, opened 2021-05-11
Update runtime options sample

jhuleatt:
- make snippets more compact
- show how to set config based on env vars


egilmorez (approved):
Thanks Jeff!  Let's ship it.

PR #945, opened 2022-01-14
fix: last sign in time

Veargan:
I've changed detection mechanism of inactive users because `user.metadata.lastSignInTime` returns the date the user last signed in and only updated when user manually logs out and signs in again. That leads to deletion of active users. So `user.metadata.lastRefreshTime` should be used instead, it returns the time at which the user was last active but could be null if user wasn't active at all, so we still should get `user.metadata.lastSignInTime`. Related issue #829


jhuleatt (approved):

jhuleatt (approved):

jhuleatt:
Thank you @Veargan!

PR #958, opened 2022-02-20
add manage FCM tokens sample

kroikie:


puf:
Nice! 

I often refer to this [code in the FCM sample](https://github.com/firebase/functions-samples/blob/main/fcm-notifications/functions/index.js#L87-L108) on how to prune tokens while sending. Is it necessary to use both approaches, or would one or the other suffice?

If both are needed, can you make sure we link them to each other?

egilmorez:
Not to pre-empt a response from Arthur, but I'd say that we are indeed recommending the use of multiple approaches to maintain token health.  The sample you cite Puf -- and thanks, because I think I can use it if we add some start/end tags -- pertains to this approach I think: https://firebase.google.com/docs/cloud-messaging/manage-tokens#detect-invalid-token-responses-from-the-fcm-backend

General comment for Arthur: can you add those tags if appropriate, and also tag the other two functions in your index.js?  Thanks!

fcm-manage-tokens/functions/index.js
```diff
@@ -0,0 +1,61 @@
...
+const EXPIRATION_TIME = 1000 * 60 * 60 * 24 * 30; // 30 days
+
+/**
+ * Callable function that stores/updates the freshness timestamp associated with the
+ * submitted FCM registration token.
+ *
+ * Note: Most FCM registration tokens would be associated with a user. In that case
+ * there should be some mapping between the user and the token(s) when stored.
+ */
+exports.updateToken = functions.https.onCall(async (data, context) => {
```
jhuleatt, line 32: @kroikie what do you think about making this a Firestore trigger instead of a callable? Reasoning is that then the writes to Firestore can be protected by Security Rules

PR #963, opened 2022-04-21
porting https trigger

racheldosh:
Porting HTTPS trigger. 

Open question: should I include a license/copyright statement?

### DIFF
--- a/2nd-gen/time-server/README.md
+++ b/2nd-gen/time-server/README.md
@@ -1,0 +1,103 @@
+# Firebase SDK for Cloud Functions Quickstart - HTTPS trigger
+
+This quickstart demonstrates using the **Firebase SDK for Cloud Functions** with an HTTPS trigger through building an endpoint returning the current time.
+
+
+## Introduction
+
+The function `date` returns the current server date. You can pass it a `format` URL Query parameter to format the date.
+
+Further reading:
+
+ - [Read more about the Firebase SDK for Cloud Functions](https://firebase.google.com/docs/functions)
+
+
+## Initial setup, build tools and dependencies
+
+### 1. Clone this repo
+
+Clone or download this repo and open the `quickstarts/time-server` directory.
+
+
+### 2. Create a Firebase project and configure the quickstart
+
+Create a Firebase Project on the [Firebase Console](https://console.firebase.google.com).
+
+Set up your Firebase project by running `firebase use --add`, select your Project ID and follow the instructions.
+
+
+### 3. Install the Firebase CLI and enable Functions on your Firebase CLI
+
+You need to have installed the Firebase CLI. If you haven't run:
+
+```bash
+npm install -g firebase-tools
+```
+
+> Doesn't work? You may need to [change npm permissions](https://docs.npmjs.com/getting-started/fixing-npm-permissions).
+
+
+## Deploy the app to prod
+
+First you need to install the `npm` dependencies of the functions:
+
+```bash
+cd functions && npm install; cd ..
+```
+
+This installs locally:
+ - The Firebase SDK and the Firebase Functions SDK.
+ - The [moment](https://www.npmjs.com/package/moment) npm package to format time.
+
+Deploy to Firebase using the following command:
+
+```bash
+firebase deploy
+```
+
+This deploys and activates the date Function.
+
+> The first time you call `firebase deploy` on a new project with Functions will take longer than usual.
+
+
+Alteratively, you can call `firebase emulators:start` to test the functions on the local emulator suite.
+
+
+## Try the sample
+
+After deploying the function you can open the following URLs in your browser (a separate instance is created for each region):
+
+```
+https://us-central3-<project-id>.cloudfunctions.net/date
+
+https://us-central3-<project-id>.cloudfunctions.net/date?format=MMMM%20Do%20YYYY%2C%20h%3Amm%3Ass%20a
+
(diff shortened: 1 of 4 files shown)


jhuleatt (changes requested):
Thank you for porting the [v1 time server sample](https://github.com/firebase/functions-samples/tree/main/quickstarts/time-server) @racheldosh! I left a few comments but overall I think this is ready for broader review.

2nd-gen/time-server/functions/index.js
```diff
@@ -0,0 +1,60 @@
+/**
+ * TODO: license
+ */
+'use strict';
+
+// [START functionsimport]
+const functions = require('firebase-functions/v2');
```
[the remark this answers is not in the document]
jhuleatt: note: you'll need to update the spots where you call `onRequest` or use the logger, too.

2nd-gen/time-server/README.md
```diff
@@ -0,0 +1,103 @@
...
+## Introduction
+
+The function `date` returns the current server date. You can pass it a `format` URL Query parameter to format the date.
+
+Further reading:
+
+ - [Read more about the Firebase SDK for Cloud Functions](https://firebase.google.com/docs/functions)
+
+
+## Initial setup, build tools and dependencies
```
jhuleatt, line 15: @egilmorez maybe we should pare this down so we're not duplicating docs? This probably applies to a lot of samples across the entire repo and we won't change all those now, but we can at least be consistent in v2 samples.

egilmorez: This does seem lengthy . . . but I wouldn't want to do anything non-essential to slow the wheels of progress, and this feels like a nice-to-have.

2nd-gen/time-server/functions/index.js
```diff
@@ -0,0 +1,60 @@
+/**
+ * TODO: license
+ */
+'use strict';
+
+// [START functionsimport]
```
egilmorez, line 6: I *think* we decided to prepend all the v2 tag names with "v2," is that correct Jeff?

So this one would be [START v2functionsimport]

It definitely will be helpful if the snippet CLI passes that naming into the final snippet name, which I think it will, and will test when the first v2 sample gets merged.

egilmorez (commented):
So happy to see this started, thanks Rachel!

2nd-gen/time-server/functions/index.js
jhuleatt: yes, let's prepend with v2

egilmorez (approved):
This looks great!  racheldosh matched the granularity of the v1 tags perfectly.

2nd-gen/time-server/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+```
+https://us-central3-<project-id>.cloudfunctions.net/date
+
+https://us-central3-<project-id>.cloudfunctions.net/date?format=MMMM%20Do%20YYYY%2C%20h%3Amm%3Ass%20a
+
+https://us-west1-<project-id>.cloudfunctions.net/date
+
+https://us-west1-<project-id>.cloudfunctions.net/date?format=MMMM%20Do%20YYYY%2C%20h%3Amm%3Ass%20a
+```
```
jhuleatt, line 78:
```suggestion
After deploying the function, check the CLI's output to see the URL for your function.

It will look something like: `https://<function-name>-<random-hash>-<region>.a.run.app`
```

CF3v2 URLs aren't deterministic yet, so we have to be a little more vague

2nd-gen/time-server/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+You can also send the format in the request body. For instance using cURL in the command line:
+
+```bash
+curl -H 'Content-Type: application/json' /
+     -d '{"format": "MMMM Do YYYY, h:mm:ss a"}' /
+     https://us-central3-<project-id>.cloudfunctions.net/date
```
jhuleatt, line 85:
```suggestion
     <function url>
```

2nd-gen/time-server/README.md
```diff
@@ -0,0 +1,103 @@
...
+```
+Formatted dates should be displayed.
+
+We are responding with a 403 error in case of PUT requests:
+
+```bash
+curl -X PUT -d '{"format": "MMMM Do YYYY, h:mm:ss a"}' https://us-central3-<project-id>.cloudfunctions.net/date
```
jhuleatt, line 92:
```suggestion
curl -X PUT -d '{"format": "MMMM Do YYYY, h:mm:ss a"}' <function-url>/date
```

jhuleatt (changes requested):
Overall this is looking great, just a couple of required changes before we can merge!

egilmorez:
Actually Rachel, I just  realized: we have a "usingMiddleware" snippet name in the v1 version.  Is there an equivalent here, or did you have to omit middleware code?

jhuleatt:
> Actually Rachel, I just realized: we have a "usingMiddleware" snippet name in the v1 version. Is there an equivalent here, or did you have to omit middleware code?

Summarizing conclusion from chat messages: CORS is handled in the options passed to the function now, so we're going to leave it out of this sample. May come back and add a different middleware example in the future.

jhuleatt (approved):
Thanks @racheldosh and @rachelmyers!

PR #973, opened 2022-04-28
Add sample for function codebases

taeold:
Simple demonstration for using codebase feature to setup a multi-codebase project repository.


quickstarts/multicodebase-hellos/README.md
```diff
@@ -0,0 +1,49 @@
+# Firebase SDK for Cloud Functions Quickstart - Multi-codebases
+
+This quickstart demonstrates how to setup a multi-codebase repository that allows you to setup multiple source directories to manage your function definitions.
```
egilmorez, line 3: When it's a verb, it needs the space: "set up," here and the next instance (the following heading is fine as it's presumably a noun).

Important internal reference: https://memegen.corp.google.com/6809714

quickstarts/multicodebase-hellos/README.md
```diff
@@ -0,0 +1,49 @@
...
+npm install -g firebase-tools@latest
+```
+
+Deploy the sample functions using the Firebase CLI deploy command:
+
+```bash
+firebase deploy --only functions
+```
+
+You can also deploy functions from specific codebase:
```
egilmorez, line 44: Suggest "a specific codebase"

quickstarts/multicodebase-hellos/README.md
```diff
@@ -0,0 +1,49 @@
+# Firebase SDK for Cloud Functions Quickstart - Multi-codebases
+
+This quickstart demonstrates how to setup a multi-codebase repository that allows you to setup multiple source directories to manage your function definitions. 
+
+
+## Setup
+
+This repository contains 2 different function sources:
```
egilmorez, line 8: Will a codebase always correspond to a language, or is it more arbitrary?  

We may not need to talk about that here, but definitely would in the docs.

egilmorez (approved):
LG!  A few nits and a question.

quickstarts/multicodebase-hellos/README.md
taeold: It can be in the same language.

For example, say you have a function that requires a heavy initialization which affects the cold start time. We've previously recommended that users do lazy initialization (https://firebase.google.com/docs/functions/tips#do_lazy_initialization_of_global_variables) so the cost of initialization doesn't affect other functions in the file.

With codebases, you can separate out the function into different codebases so you can group functions that requires the heavy initialization (w/ corresponding impact on cold start) and ones that does not!

PR #975, opened 2022-05-02
Create sample for unit testing gen2 functions

TheIronDev:
Creating a sample for unit testing gen2 functions.


thatfiredev (approved):
LGTM

PR #985, opened 2022-05-11
Add information about v2 to readme files

jhuleatt:


README.md
```diff
@@ -18,6 +18,7 @@ To learn how to get started with Cloud Functions for Firebase by having a look a
 This repository contains a few categories of samples:
 
  1. [**Quickstarts**](#quickstarts) are minimal examples for each types of triggers.
+ 1. [**Cloud Functions v2 public preview**](#v2) samples show how to use the public preview of [Cloud Functions for Firebase v2](https://firebase.google.com/docs/functions/beta).
  1. [**Development Environment Samples and Boilerplates**](#environment) illustrates how to get started with
 different, commonly used JavaScript development patterns such as Typescript, React SSR, ES2017 etc...
  1. [**Image Processing**](#image) shows a few ways where you can process and transform images using Cloud Functions such as generating thumbnails, converting images extracting metadata...
```
egilmorez, line 24: This could just be "a few ways you can process"

README.md
```diff
@@ -18,6 +18,7 @@ To learn how to get started with Cloud Functions for Firebase by having a look a
 This repository contains a few categories of samples:
 
  1. [**Quickstarts**](#quickstarts) are minimal examples for each types of triggers.
+ 1. [**Cloud Functions v2 public preview**](#v2) samples show how to use the public preview of [Cloud Functions for Firebase v2](https://firebase.google.com/docs/functions/beta).
  1. [**Development Environment Samples and Boilerplates**](#environment) illustrates how to get started with
 different, commonly used JavaScript development patterns such as Typescript, React SSR, ES2017 etc...
```
egilmorez, line 23: Suggest "SSR, and ES2017."

I'd do the same with the next two ellipses below.  "Such as" lets you give a few examples, but suggests more.

egilmorez (approved):
A couple of thoughts, possibly out of scope :)

Confused by the diff; I can't see the instances of "usecases" in the markdown -- those should be "use cases," and it would be nice to lead into that chunk with a verb like the other items.

v2 material all LG, thanks!

PR #996, opened 2022-05-29
Fix for delete-unused-accounts-cron example

edward62740:
Running this example results in TypeError: PromisePool is not a constructor.

Changed
``const PromisePool = require('es6-promise-pool').default;``
to
``const PromisePool = require('es6-promise-pool');``

This change fixes the abovementioned error.

### DIFF
--- a/delete-unused-accounts-cron/functions/index.js
+++ b/delete-unused-accounts-cron/functions/index.js
@@ -18,7 +18,7 @@
 const functions = require('firebase-functions');
 const admin = require('firebase-admin');
 admin.initializeApp();
-const PromisePool = require('es6-promise-pool').default;
+const PromisePool = require('es6-promise-pool');
 // Maximum concurrent account deletions.
 const MAX_CONCURRENT = 3;
 


jhuleatt:
Thanks @edward62740!

kaibolay:
@jhuleatt This seems to have broken our CI tests ([example](https://github.com/firebase/functions-samples/actions/runs/3107917446/jobs/5036546937)):

```
Error: delete-unused-accounts-cron/functions compile: index.js(33,27): error TS2351: This expression is not constructable.
delete-unused-accounts-cron/functions compile:   Type 'typeof import("/home/runner/work/functions-samples/functions-samples/node_modules/.pnpm/es6-promise-pool@2.5.0/node_modules/es6-promise-pool/es6-promise-pool")' has no construct signatures.
delete-unused-accounts-cron/functions compile: Failed
```

1. Any idea how to fix that without reverting this PR?
2. Any idea why the CI tests didn't run for this PR (which would presumably have caught this problem)?

PR #1010, opened 2022-08-29
Fixed onDisconnect logic to use deviceId

maneesht:


presence-firestore/functions/index.js
```diff
@@ -47,11 +51,20 @@ exports.onUserStatusChanged = functions.database.ref('/status/{uid}').onUpdate(
...
       // Otherwise, we convert the last_changed field to a Date
       eventStatus.last_changed = new Date(eventStatus.last_changed);
 
-      // ... and write it to Firestore.
-      return userStatusFirestoreRef.set(eventStatus);
+      if(status.state === 'offline') {
+        // ... and write it to Firestore.
+        await userStatusFirestoreRef.delete();
+
+        if ((await userDeviceCollectionRef.get()).empty) {
```
maneesht, line 61: `async/await` syntax targets ES2016+, which is inconsistent with the rest of the file. Looking for whether ES5 is the target, or ES2016+

presence-firestore/public/index.js
```diff
@@ -21,7 +21,8 @@ function rtdb_presence() {
 
     // Create a reference to this user's specific status node.
     // This is where we will store data about being online/offline.
-    var userStatusDatabaseRef = firebase.database().ref('/status/' + uid);
+    // `push()` will add a new key to the user's path, acting as a deviceId
+    var deviceStatusDatabaseRef = firebase.database().ref('/status/' + uid + '/devices').push();
```
jsdt, line 25: `sessions` or `connections` is probably more accurate than `devices`, since the id isn't based on the device id.

PR #1012, opened 2022-09-16
Add example code for Performance threshold alert

jfcong:
Add example code for Performance threshold alert for 2nd gen function alert to discord.

### DIFF
--- a/2nd-gen/alerts-to-discord/functions/index.js
+++ b/2nd-gen/alerts-to-discord/functions/index.js
@@ -183,7 +183,7 @@ exports.postperformancealerttodiscord = onThresholdAlertPublished(
       // [END v2PerformanceEventPayload]
 
       try {
-        // [START v2SendToDiscord]
+        // [START v2SendPerformanceAlertToDiscord]
         const response = await postMessageToDiscord(
             "Firebase Performance Bot", message);
         if (response.ok) {
@@ -194,7 +194,7 @@ exports.postperformancealerttodiscord = onThresholdAlertPublished(
         } else {
           throw new Error(response.error);
         }
-        // [END v2SendToDiscord]
+        // [END v2SendPerformanceAlertToDiscord]
       } catch (error) {
         logger.error(
             `Unable to post Firebase Performance alert ${eventName} to Discord`,
--- a/2nd-gen/alerts-to-discord/functions/index.js
+++ b/2nd-gen/alerts-to-discord/functions/index.js
@@ -21,6 +21,9 @@ const {
 const {
   onNewTesterIosDevicePublished,
 } = require("firebase-functions/v2/alerts/appDistribution");
+const {
+  onThresholdAlertPublished,
+} = require("firebase-functions/v2/alerts/performance");
 const logger = require("firebase-functions/logger");
 // [END v2import]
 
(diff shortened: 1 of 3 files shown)


2nd-gen/alerts-to-discord/functions/index.js
```diff
@@ -86,9 +94,112 @@ id: \`${id}\`
...
+    
+    **Investigate more:** ${investigateUri}
+    `;
+      // [END v2PerformanceEventPayload]
+
+      try {
+        // [START v2SendToDiscord]
+        const response = await postMessageToDiscord(
+            "Firebase Performance Bot", message);
+        if (response.ok) {
```
kaibolay, line 189: Do you think we should pull this code (probably also the `try`/`catch`) into `postMessageToDiscord()`? It's not replicated three times with only minimal differences in the info/error message being logged. We could just make the error messages more generic.

The complication is that this error handling is currently in the snippets (`[START v2SendToDiscord]`/`[END v2SendToDiscord]`)...

jfcong: I discussed this with Victoria, I think what we can do is leave this as is, but in the documentation, just link to the code snippet without showing the code itself, so it doesn't look as redundant in the doc. wdyt?

Ideally we can clean this up in the code snippets as well but as you said, it's currently used in the doc...

2nd-gen/alerts-to-discord/functions/index.js
```diff
@@ -86,9 +94,112 @@ id: \`${id}\`
...
+    
+    Violation: ${violationValue} ${violationUnit}
+    Number of samples checked: ${numSamples}
+    
+    **Investigate more:** ${investigateUri}
+    `;
+      // [END v2PerformanceEventPayload]
+
+      try {
+        // [START v2SendToDiscord]
```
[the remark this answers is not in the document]
jfcong: Thanks for the catch! Updated the tag. (ideally we can update it for crash too...)

jhuleatt:
Thanks for the performance sample @jfcong! I'm a bit confused though - how does this PR relate to https://github.com/firebase/functions-samples/pull/1012? Seems like both add the iOS app distro sample. Does it replace that one?

2nd-gen/alerts-to-discord/functions/index.js
kaibolay: Ok. We can always to the de-duplication in a follow-up.

kaibolay:
> Thanks for the performance sample @jfcong! I'm a bit confused though - how does this PR relate to #1012? Seems like both add the iOS app distro sample. Does it replace that one?

@jhuleatt Jenny's PR builds on top of mine (#1012), so it also includes those changes.

You can look at [only her changes](https://github.com/firebase/functions-samples/pull/1012/files/1b2d4eefa86bf11010927240981c04f7b1ffad66..f7ca5537cf72d23d54895efe82d474ea3f980ef5) by selecting only her commits for the diff.

The plan is that I'll first merge my PR, then Jenny will rebase her branch (so it only includes her changes), update this PR, and then merge it.

jfcong:
@jhuleatt Like Kai said, I will rebase once Kai's PR is submitted :)

Thanks Kai for replying!

kaibolay (approved):

jhuleatt (approved):
LGTM, please get tech writer approval of the includecode snippets before merging. Thanks again for contributing a sample!

2nd-gen/alerts-to-discord/functions/index.js
```diff
@@ -139,3 +142,64 @@ UDID **${testerDeviceIdentifier}** for ${testerDeviceModelName}
...
+        violationUnit,
+        investigateUri,
+      } = event.data.payload;
+      const message = `
+    ⚠️ Performance Alert for ${metricType} of ${eventType}: **${eventName}** ⚠️
+    
+    App id: ${appId}
+    Alert condition: ${thresholdValue} ${thresholdUnit}
+    Percentile (if applicable): ${conditionPercentile}
+    App version (if applicable): ${appVersion}
```
jhuleatt, line 176: nit: consider making this conditional instead of having the "if applicable" qualifier

2nd-gen/alerts-to-discord/functions/index.js
```diff
@@ -139,3 +142,64 @@ UDID **${testerDeviceIdentifier}** for ${testerDeviceModelName}
...
+    
+    Violation: ${violationValue} ${violationUnit}
```
jhuleatt, line 178: This seems like the most important thing - consider moving it higher up in the message

jfcong: I grouped it by two clusters, the first cluster is alert condition, and the second cluster is the alert violation. This resembles how our alert email displays the information (with a focus on the alert condition as well). I'll let a tech writer provide some input for this one :)

2nd-gen/alerts-to-discord/functions/index.js
jfcong: I am wondering what you mean by conditional, do you mean having 4 if statements and depending on which one is true, having a different string message? I am leaning towards just keeping it simple because this is just an example for developers and they should be able to format the message however they want.

2nd-gen/alerts-to-discord/functions/index.js
```diff
@@ -139,3 +142,64 @@ UDID **${testerDeviceIdentifier}** for ${testerDeviceModelName}
   }
 });
 // [END v2Alerts]
+
+/**
+ * Function triggered by Firebase Performance that publishes
```
VictoriaAuto, line 147: Update product name to "Firebase Performance Monitoring"

2nd-gen/alerts-to-discord/functions/index.js
VictoriaAuto: To provide a consistent experience across surfaces, I recommend keeping the current flow. Also, the example isn't that long, and a violation is implied in the description; I think developers will expect that it's a part of the example and will look for it.

jfcong: sounds good, I'll keep it as is

2nd-gen/alerts-to-discord/functions/index.js
jfcong: Updated.

VictoriaAuto (approved):

2nd-gen/alerts-to-discord/functions/index.js
jfcong: I am going to keep this simple and not do conditional, since this is meant to just be some sample code

PR #1022, opened 2022-10-25
Fix bugs and add misising files in the otel sample.

taeold:


2nd-gen/instrument-with-opentelemetry/functions/.env
```diff
@@ -0,0 +1 @@
+NODE_OPTIONS=--require=./tracing.js
```
jhuleatt, line 1: is this needed?

taeold: yes - it's the only way I've found to load otel libraries before functions framework is loaded. Due to how monkeypatching works, otel sdk has to be the very first thing that runs before functions framework or user code is loaed.

jhuleatt, line 1:
```suggestion
# This loads the telemetry library before functions framework because
# The Open Telemetry SDK must be the first thing that runs
NODE_OPTIONS=--require=./tracing.js
```

jhuleatt (approved):

2nd-gen/instrument-with-opentelemetry/functions/.env
jhuleatt: Interesting, I didn't know we could do this. Thanks for explaining!

jhuleatt:
Thanks @taeold!

PR #1045, opened 2023-02-24
Fix vision-annotate-images sample

colerogers:
Fixes https://github.com/firebase/firebase-tools/issues/5502


jhuleatt:
Thanks @colerogers!

PR #1051, opened 2023-03-10
Add pubsub "Hello World" sample

kevinthecheung:


Python/quickstarts/pubsub-helloworld/main.py
```diff
@@ -0,0 +1,77 @@
...
+    # [START readJson]
+    # Get the `name` attribute of the PubSub message JSON body.
+    try:
+        data = event.data.message.json
+    except ValueError:
+        print("PubSub message was not JSON")
+        return
+    # [END readJson]
+
+    if "name" not in data:
```
jhuleatt, line 55: can you access `data` if it is only defined in the `try` block?

jhuleatt (approved):

Python/quickstarts/pubsub-helloworld/main.py
kevinthecheung: Yes, `try` doesn't create a new scope

PR #1062, opened 2023-04-06
Python Cloud Tasks sample

kevinthecheung:


egilmorez (approved):
LGTM, thanks!

Python/taskqueues-backup-images/main.py
```diff
@@ -0,0 +1,158 @@
...
+        )
+        tasks_client.create_task(parent=task_queue, task=task)
+
+    return https_fn.Response(status=200,
+                             response=f"Enqueued {BACKUP_COUNT} tasks")
+# [END v2EnqueueTasks]
+
+
+# [START v2GetFunctionUri]
+def get_function_url(name: str,
```
jhuleatt, line 139: Now that cloudfunctions.net URLs are available in gen2, we can remove this from both the Node and Python sample. But it's still being tested in CF3 so it's not worth blocking this PR on that

jhuleatt (approved):
Thanks Kevin!

PR #1069, opened 2023-04-14
Test Lab Python sample

kevinthecheung:


kevinthecheung:
@jhuleatt I have not tested this.

Python/testlab-to-slack/main.py
```diff
@@ -0,0 +1,99 @@
...
+# The requests library to send web requests to Slack.
+import requests
+# [END import]
+
+
+# [START postToSlack]
+def post_to_slack(title: str, details: str) -> any:
+    """Posts a message to Slack via a Webhook."""
+    response = requests.post(
+        url=os.environ["SLACK_WEBHOOK_URL"],
```
jhuleatt, line 31: Can you please use params for this? https://github.com/firebase/firebase-functions-python/pull/2/files

We should eventually have it as a Secret, but secret params aren't supported in Python yet

jhuleatt: hm, actually the Node sample uses `process.env` at the moment as well. Changing to params will probably change the narrative in the docs, so let's stick to env vars as you have it now

jhuleatt (approved):
Thank you @kevinthecheung

jhuleatt (changes requested):
Let's make sure we get a chance to test this before merging

kevinthecheung:
> Let's make sure we get a chance to test this before merging

Tested, and it works.

PR #1070, opened 2023-04-15
Scheduler Python sample

kevinthecheung:


kevinthecheung:
@jhuleatt I think this is adequate for Eric to work on docs. I can write a more scalable version with Cloud Tasks later if desirable.

egilmorez (approved):
LGTM!

RE the better example with Cloud Tasks: from a high level, it seems as though a robust Cloud Tasks capability might completely obsolete (v.) this approach.

Same question on whether logging needs to be imported separately.  Thanks!

Python/delete-unused-accounts-cron/main.py
```diff
@@ -0,0 +1,61 @@
...
+
+def is_inactive(user: auth.UserRecord, inactive_limit: timedelta) -> bool:
+    if user.user_metadata.last_refresh_timestamp is not None:
+        last_seen_timestamp = user.user_metadata.last_refresh_timestamp / 1000
+    elif user.user_metadata.last_sign_in_timestamp is not None:
+        last_seen_timestamp = user.user_metadata.last_sign_in_timestamp / 1000
+    elif user.user_metadata.creation_timestamp is not None:
+        last_seen_timestamp = user.user_metadata.creation_timestamp / 1000
+    else:
+        # Impossible?
```
jhuleatt, line 56: nit: no need for this comment

```suggestion
```

jhuleatt (approved):

PR #1072, opened 2023-04-22
Eventarc Python sample

kevinthecheung:


egilmorez (approved):
This looks great, thanks!

I'm assuming that we may or may not have Pythonic snippets for creating channels, and that those would be either out of scope for this sample or in a different file than main.py.

kevinthecheung:
> This looks great, thanks!
> 
> I'm assuming that we may or may not have Pythonic snippets for creating channels, and that those would be either out of scope for this sample or in a different file than main.py.

Yeah, that's an Admin SDK snippet, so we don't necessarily need a Python example. Not here regardless.

But, I did add a snippet just now for non-default channels.

jhuleatt (approved):

jhuleatt:
Thanks @kevinthecheung!

PR #1073, opened 2023-04-24
Blocking auth Python samples

kevinthecheung:


Python/auth-blocking-functions/main.py
```diff
@@ -0,0 +1,231 @@
...
+            message="You must register using a trusted provider.",
+        )
+# [END requireverified]
+
+
+def send_verification_email_using_your_smtp_server(email, link):
+    return
+
+
+# TODO: Should really be non-blocking or client-side call.
```
kevinthecheung, line 95: I don't think we should include this example.

1. There's no need for this to be blocking, AFAIK.
2. The (Python) Admin SDK intentionally doesn't support sending emails from Firebase, so there's a big hand-wave.

kevinthecheung: I think we should remove or rework this example altogether, for both Python and Node, for the reasons given above.

egilmorez: Roger that!  Saw your other PR (thumbs up emoji).

jhuleatt:
Hey @kevinthecheung, should this PR be closed in favor of #1074, or should I review this too?

kevinthecheung:
> Hey @kevinthecheung, should this PR be closed in favor of #1074, or should I review this too?

Please review.

The docs use snippets from both. I just split off #1074 because it has several moving parts.

egilmorez (approved):
Lets get these merged, thanks

Python/auth-blocking-functions/main.py
```diff
@@ -0,0 +1,231 @@
...
+    # [END v2beforeSignInFunctionTrigger]
+    # [START v2readEmailData]
+    # Email passed from the CloudEvent.
+    email = event.data.email if event.data.email is not None else ""
+    # [END v2readEmailData]
+
+    # [START v2documentGet]
+    # Obtain a document in Firestore of the banned email address.
+    firestore_client: google.cloud.firestore.Client = firestore.client()
+    doc = firestore_client.collection("banned").document(email).get()
```
jhuleatt, line 219: Can we use the Firebase Admin SDK here instead?

kevinthecheung: We do, that's just for typing/Intellisense

PR #1079, opened 2023-05-04
Remove ImageMagick from thumbnail samples (1st and 2nd gen)

jhuleatt:
context: https://github.com/firebase/firebase-tools/issues/5489


egilmorez (approved):
Hooray for sharp!!  Thanks!

egilmorez (approved):
Thanks for extending to 1st gen as well!

PR #1132, opened 2024-04-05
Add samples for Firestore with Auth context

jhuleatt:


Node/quickstarts/firestore-sync-auth/functions/index.js
```diff
@@ -0,0 +1,61 @@
...
+        verified = true;
+      } else if (authType === "unknown" || authType === "unauthenticated") {
+      // admin users from a specific domain are verified
+        if (authId.endsWith("@example.com")) {
+          verified = true;
+        }
+      }
+
+      // add auth medadata to the document
+      return data.after.ref.set(
```
blidd-google, line 51: I believe this needs to be `snapshot.ref.set(...)` instead of `data.after.ref.set(...)`

Node/quickstarts/firestore-sync-auth/functions/index.js
```diff
@@ -0,0 +1,61 @@
...
+          {
+            created_by: authId,
```
blidd-google, line 53: In case authId is undefined, we need: `created_by: authId ?? "undefined"`, since Firestore doesn't allow using `undefined` as a property value by default.

blidd-google (approved):
LGTM besides some minor fixes

egilmorez (approved):
Thanks!

PR #1134, opened 2024-04-26
feat: add call gemini for v2

HYACCCINT:


Node/call-gemini/functions/index.js
```diff
@@ -0,0 +1,39 @@
...
+ */
+ const {onCall, HttpsError} = require("firebase-functions/v2/https");
+ const {logger} = require("firebase-functions/v2");
+ const { GoogleGenerativeAI } = require("@google/generative-ai");
+ const { defineSecret } = require('firebase-functions/params');
+
+ const geminiToken = defineSecret("API_TOKEN", {
+  description: "Gemini API token. Created using " +
+      "https://ai.google.dev/tutorials/get_started_node#set-up-project",
+});
```
jhuleatt, line 24: this is great! thanks for adding a descriptive prompt.

Node/call-gemini/functions/index.js
```diff
@@ -0,0 +1,39 @@
...
+
+ exports.callGemini = onCall({secrets: [geminiToken]}, async (request) => {
+
```
jhuleatt, line 27: Before calling Gemini, let's look for both appcheck and auth, but gate each with its own boolean parameter (https://firebase.google.com/docs/functions/config-env?gen=2nd#parameter_types) so devs can choose whether they want auth/app check or not.

Something like:

```js
// ...imports

const appCheckRequired = defineBoolean("APP_CHECK_REQUIRED");
const authRequired = defineBoolean("AUTH_REQUIRED");

// ...other setup

exports.callGemini = onCall(
  {
    enforceAppCheck: appCheckRequired, // Reject requests with missing or invalid App Check tokens.
  },
  (request) => {
    if (authRequired.val() && !request.auth) {
      throw new HttpsError(
        "failed-precondition",
        "The function must be called while authenticated.",
      );
    }

    // ...call Gemini
  },
);
```

package.json
```diff
@@ -25,5 +25,8 @@
         "@firebase/app-types"
       ]
     }
+  },
+  "dependencies": {
+    "firebase": "^10.11.1"
```
jhuleatt, line 30: I think this was an accident. Might mean root-level `package-lock.json` update was too.

pnpm-workspace.yaml
```diff
@@ -15,3 +15,4 @@
 packages:
   - 'Node/**'
   - 'Node-1st-gen/**'
+  - "!Node-1st-gen/image-maker/**"
```
jhuleatt, line 18: please remove before merging

Node/call-gemini/functions/package.json
```diff
@@ -0,0 +1,30 @@
...
+  },
+  "engines": {
+    "node": "18"
+  },
+  "main": "index.js",
+  "dependencies": {
+    "@google/generative-ai": "^0.5.0",
+    "firebase-admin": "^11.9.0",
+    "firebase-functions": "^4.4.1",
+    "node-fetch": "^2.6.7"
```
jhuleatt, line 21: is `node-fetch` needed for this sample?

PR #1135, opened 2024-05-22
Add function sample for server-side Remote Config + Vertex AI Gemini API

jenh:
Adds the function and snippets to support https://firebase.google.com/docs/remote-config/solution-server


Node/remote-config-server-with-vertex/functions/index.js
```diff
@@ -0,0 +1,119 @@
...
+    const vertexEnabled = config.getBoolean("is_vertex_enabled") ||
+        defaultConfig.vertex_enabled;
+// [END remote_config_server_vertex_create_function]
+
+// [START remote_config_server_vertex_function_logic]
+  // Allow user input.
+  const userInput = request.query.prompt || '';
+
+  // Instantiate Vertex AI.
+    const vertex_ai = new VertexAI({ project: project, location: location });
```
jhuleatt, line 76: Are project and location needed, or is that covered by [Application Default Credentials](https://cloud.google.com/docs/authentication/application-default-credentials)

Node/remote-config-server-with-vertex/functions/index.js
```diff
@@ -0,0 +1,119 @@
...
+  // Disable Vertex AI Gemini API access for testing.
+  vertex_enabled: false
+};
+// [END remote_config_server_vertex_default_values]
+
+// [START remote_config_server_vertex_create_function]
+// Export the function.
+exports.generateWithVertex = onRequest(async (request, response) => {
+
+  try {
```
jhuleatt, line 44: It would be great to add some kind of auth check here to encourage developers to protect their AI endpoints. https://github.com/firebase/functions-samples/pull/1134#discussion_r1583181807 does that with `onCall` + AppCheck + Auth, but this could check Auth with something like: https://firebase.google.com/docs/auth/web/service-worker-sessions#server_side_changes

Node/remote-config-server-with-vertex/functions/index.js
```diff
@@ -0,0 +1,119 @@
...
+  // Instantiate Vertex AI.
+    const vertex_ai = new VertexAI({ project: project, location: location });
+    const generativeModel = vertex_ai.getGenerativeModel({
+      model: textModel,
+      safety_settings: safetySettings,
+      generation_config: generationConfig,
+    });
+
+    // Create the chat; append user input to Remote Config-defined prompt.
+    const chat = generativeModel.startChat();
```
jhuleatt, line 84: if it's a single call per request, with no history, is [`generateContentStream`](https://github.com/googleapis/nodejs-vertexai?tab=readme-ov-file#get-streamed-text-responses) more appropriate than `startChat`? I haven't used Vertex much so I don't really know.

Node/remote-config-server-with-vertex/functions/index.js
```diff
@@ -0,0 +1,119 @@
...
+    }
+    // If vertexEnabled isn't true, do not send queries to Vertex AI.
+    if (vertexEnabled !== true) {
+      response.status(200).send({
+        message: "Vertex AI call skipped. Vertex is not enabled."
+      });
+      return;
+    }
+
+    console.log("\nRunning with model ", textModel, ", prompt: ", textPrompt,
```
jhuleatt, line 98: This can be a structured log with the logger sdk: https://firebase.google.com/docs/functions/writing-and-viewing-logs?gen=2nd

Node/remote-config-server-with-vertex/functions/index.js
jhuleatt: Oh, you do have the `vertex_enabled` flag already. Auth would be ideal, but that flag is probably ok for now

Node/remote-config-server-with-vertex/functions/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "functions",
```
jhuleatt, line 2:
```suggestion
  "name": "remote-config-server-with-vertex",
```

jhuleatt (approved):
Thank you for adding a Vertex sample, Jen!

Node/remote-config-server-with-vertex/functions/index.js
jenh: But the likelihood of folks leaving vertex_enabled to true exists...so I will get some auth in here! I think the service-worker-sessions method would actually work really well with the tutorial, so I'll try this out.

Node/remote-config-server-with-vertex/functions/index.js
jenh: Yep, that is better and makes it simpler! Updated this here and in the callable function.

Node/remote-config-server-with-vertex/functions/index.js
jenh: Yep, looks like they are needed and it doesn't automatically pull ADC. :/ Thinking it's because you can use Vertex in a different locale than your project.

thatfiredev (approved):

Node/call-vertex-remote-config-server/client/main.ts
```diff
@@ -0,0 +1,118 @@
...
+
+promptForm.addEventListener("submit", async (event) => {
+  event.preventDefault();
+  const promptInput = document.getElementById(
+    "promptInput"
+  ) as HTMLInputElement;
+  const prompt = promptInput.value;
+
+  const waitingMessageElement = document.getElementById("waitingMessage");
+  //waitingMessageElement.textContent = "Waiting for response...";
```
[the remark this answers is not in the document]
jenh: Thanks for catching this! Fixed! :D

HYACCCINT (approved):
Looks great, just a small nit about the node version -- 18 is approaching end of life so was wondering if it'd make sense to bump it to 20?

jenh:
> Looks great, just a small nit about the node version -- 18 is approaching end of life so was wondering if it'd make sense to bump it to 20?

I think it would...*but* it looks like we still reference 18 in the other function samples, and the basic template we create when we [init functions with the Firebase CLI](https://github.com/firebase/firebase-tools/blob/ce5d7823b6cbac6d0fbe7e7f6fcaf421700c075a/templates/init/functions/javascript/package.nolint.json#L11-L13) is still 18.

Do you think it makes sense to bump the version in separate PRs that update 18 to 20 everywhere (and adds [22 to the CLI, as it's in preview](https://cloud.google.com/functions/docs/release-notes#May_17_2024)) as a (bigger) separate effort?

HYACCCINT:
> > Looks great, just a small nit about the node version -- 18 is approaching end of life so was wondering if it'd make sense to bump it to 20?
> 
> I think it would..._but_ it looks like we still reference 18 in the other function samples, and the basic template we create when we [init functions with the Firebase CLI](https://github.com/firebase/firebase-tools/blob/ce5d7823b6cbac6d0fbe7e7f6fcaf421700c075a/templates/init/functions/javascript/package.nolint.json#L11-L13) is still 18.
> 
> Do you think it makes sense to bump the version in separate PRs that update 18 to 20 everywhere (and adds [22 to the CLI, as it's in preview](https://cloud.google.com/functions/docs/release-notes#May_17_2024)) as a (bigger) separate effort?

Oh that's a great point. In that case, absolutely updating in a separate PR would be the best path to go about it, I can look into updating the CLI as well. Thank you so much!!

jenh:
Thank YOU!!! Appreciate the review! :D

PR #1253, opened 2026-03-31
Add beforeSendEmail and beforeSendSms blocking functions samples

macastelaz:
Supplement samples in Node and Python to include examples of beforeSendEmail and beforeSendSms

### DIFF
--- a/Node/quickstarts/auth-blocking-functions/functions/index.js
+++ b/Node/quickstarts/auth-blocking-functions/functions/index.js
@@ -103,7 +103,7 @@ exports.checkEmailDomain = beforeEmailSent((event) => {
 
 const intlPrefixNumber = defineString("INTERNATIONAL_PREFIX_NUMBER", {
   default: "+1",
-  description: "The country code that we restrict sending too.",
+  description: "The country code that we restrict sending to.",
 });
 // Block SMS sending with any non-US phone number.
 exports.checkPhoneNumber = beforeSmsSent((event) => {
--- a/Node/quickstarts/auth-blocking-functions/functions/index.js
+++ b/Node/quickstarts/auth-blocking-functions/functions/index.js
@@ -29,7 +29,7 @@ const db = admin.firestore();
 // [START v2ValidateNewUser]
 // [START v2beforeCreateFunctionTrigger]
 // Block account creation with any non-acme email address.
-exports.validatenewuser = beforeUserCreated((event) => {
+exports.validateNewUser = beforeUserCreated((event) => {
   // [END v2beforeCreateFunctionTrigger]
   // [START v2readUserData]
   // User data passed in from the CloudEvent.
@@ -49,7 +49,7 @@ exports.validatenewuser = beforeUserCreated((event) => {
 // [START v2CheckForBan]
 // [START v2beforeSignInFunctionTrigger]
 // Block account sign in with any banned account.
-exports.checkforban = beforeUserSignedIn(async (event) => {
+exports.checkForBan = beforeUserSignedIn(async (event) => {
   // [END v2beforeSignInFunctionTrigger]
   // [START v2readEmailData]
   // Email passed from the CloudEvent.
@@ -74,7 +74,7 @@ exports.checkforban = beforeUserSignedIn(async (event) => {
 // [START v2CheckEmailDomain]
 // [START v2beforeEmailSentFunctionTrigger]
 // Block email sending with any non-acme email address.
-exports.checkemaildomain = beforeEmailSent((event) => {
+exports.checkEmailDomain = beforeEmailSent((event) => {
   // [END v2beforeEmailSentFunctionTrigger]
   // [START v2readEmailUser]
   // Email passed in from the CloudEvent.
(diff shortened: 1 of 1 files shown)


Node/quickstarts/auth-blocking-functions/functions/index.js
```diff
@@ -68,3 +70,44 @@ exports.checkforban = beforeUserSignedIn(async (event) => {
...
+exports.checkemaildomain = beforeEmailSent((event) => {
+  // [END v2beforeEmailSentFunctionTrigger]
+  // [START v2readEmailUser]
+  // Email passed in from the CloudEvent.
+  const email = event.data?.email || event.additionalUserInfo?.email;
+  // [END v2readEmailUser]
+
+  // [START v2emailHttpsError]
+  // Only users of a specific domain can receive emails.
+  if (!email?.includes("@acme.com")) {
```
jhuleatt: [`endsWith`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith) is probably most appropriate for this
```suggestion
  if (!email?.endsWith("@acme.com")) {
```

Node/quickstarts/auth-blocking-functions/functions/index.js
```diff
@@ -68,3 +70,44 @@ exports.checkforban = beforeUserSignedIn(async (event) => {
...
+    // Throw an HttpsError so that Firebase Auth rejects the email sending.
+    throw new HttpsError("invalid-argument", "Unauthorized email");
```
jhuleatt: What do you think about using `permission-denied` instead of `invalid-argument`?

https://firebase.google.com/docs/reference/node/firebase.functions#functionserrorcode

Node/quickstarts/auth-blocking-functions/functions/index.js
jhuleatt: instead of `email?`, consider doing a falsy check before this and throwing an `invalid-argument` exception that's separate from the "wrong email" exception

Node/quickstarts/auth-blocking-functions/functions/index.js
```diff
@@ -68,3 +70,44 @@ exports.checkforban = beforeUserSignedIn(async (event) => {
...
+  // [END v2beforeSmsSentFunctionTrigger]
+  // [START v2readSmsUser]
+  // Phone number passed from the CloudEvent.
+  const phoneNumber = event.data?.phoneNumber ||
+    event.additionalUserInfo?.phoneNumber;
+  // [END v2readSmsUser]
+
+  // [START v2smsHttpsError]
+  // Only users of a specific region can receive SMS.
+  if (!phoneNumber?.startsWith("+1")) {
```
jhuleatt: instead of `phoneNumber?`, consider doing a falsy check before this and throwing an `invalid-argument` exception that's separate from the "wrong phone number" exception

Node/quickstarts/auth-blocking-functions/functions/index.js
```diff
@@ -68,3 +70,44 @@ exports.checkforban = beforeUserSignedIn(async (event) => {
   // [END v2bannedHttpsError]
 });
 // [START v2CheckForBan]
+
+// [START v2CheckEmailDomain]
+// [START v2beforeEmailSentFunctionTrigger]
+// Block email sending with any non-acme email address.
+exports.checkemaildomain = beforeEmailSent((event) => {
```
jhuleatt: it's ok to camel case `checkEmailDomain` and `checkPhoneNumber`. I think the other functions in this file were written back when our deploy pipeline didn't support capital letters in function names

Node/quickstarts/auth-blocking-functions/functions/index.js
jhuleatt: instead of hardcoding `+1`, this sample would be more reusable if you created a [param](https://firebase.google.com/docs/functions/config-env#params). It would look something like:

```js
// outside of this function
const intlPrefixNumber = defineString("INTERNATIONAL_PREFIX_NUMBER", {
  default: "+1",
  description: "The country code that we restrict sending too.",
});
```

```js
if (!phoneNumber?.startsWith(intlPrefixNumber.value())) {
  // ...
}
```

it does make the sample a little harder to follow, though, since the param needs to be defined outside the function body. up to you

jhuleatt (approved):
Thanks @macastelaz! I added some refinement suggestions but overall LGTM

Node/quickstarts/auth-blocking-functions/functions/index.js
macastelaz: Done

Node/quickstarts/auth-blocking-functions/functions/index.js
macastelaz: Agreed! Updated

Node/quickstarts/auth-blocking-functions/functions/index.js
macastelaz: Works for me!

Node/quickstarts/auth-blocking-functions/functions/index.js
macastelaz: Done

macastelaz: Done!

Node/quickstarts/auth-blocking-functions/functions/index.js
```diff
@@ -68,3 +71,61 @@ exports.checkforban = beforeUserSignedIn(async (event) => {
...
+  // [END v2emailHttpsError]
+});
+// [END v2CheckEmailDomain]
+
+// [START v2CheckPhoneNumber]
+// [START v2beforeSmsSentFunctionTrigger]
+
+const intlPrefixNumber = defineString("INTERNATIONAL_PREFIX_NUMBER", {
+  default: "+1",
+  description: "The country code that we restrict sending too.",
```
jhuleatt:
```suggestion
  description: "The country code that we restrict sending to.",
```

jhuleatt (approved):

PR #1304, opened 2026-07-30
refactor(workflows): fix zizmor findings and improve security permissions

inlined:
### Description
- Added top-level permissions block and explanatory comments to `targeted-dependency-update.yml` to restrict default GITHUB_TOKEN permissions.
- Added top-level concurrency configuration to prevent race conditions and concurrent runs on duplicate triggers across all workflows.
- Set `persist-credentials: false` on checkouts to avoid potential credential leakage in test runs.
- Added explicit names to jobs to make results more readable in the GitHub Actions UI.

### Scenarios Tested
Ran zizmor on the modified workflows using:
- `zizmor --persona=auditor .github/workflows`
- `zizmor --pedantic --min-severity=informational .github/workflows`

Both checks passed with no findings.

### DIFF
--- a/.github/workflows/targeted-dependency-update.yml
+++ b/.github/workflows/targeted-dependency-update.yml
@@ -4,12 +4,19 @@ on:
     - cron: '0 0 * * 1' # Weekly
   workflow_dispatch: # Manual trigger
 
+concurrency:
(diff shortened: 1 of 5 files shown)


jhuleatt (approved):

jhuleatt:
Thanks!

